PR #8, opened 2017-11-01
Updates the the user in case they don't have a source

mickeytgl:
Small bug fix that updates the customer source with the token that gets passed in. This seems to be fixed in my laptop. Thoughts?

image


excid3:
@mickeytgl did you confirm that the code we were looking at yesterday is not going through the `else` of this method? And if so, why was that? I couldn't figure out off the top of my head.

lib/pay/billable/stripe.rb
```diff
@@ -3,7 +3,9 @@ module Billable
     module Stripe
       def stripe_customer
         if processor_id?
-          customer = ::Stripe::Customer.retrieve(processor_id)
+         customer = ::Stripe::Customer.retrieve(processor_id)
+         customer.source = card_token 
+         customer.save
```
excid3: Instead of setting the source, we need to call the `update_card` method, but only if the card_token is present.

https://github.com/jasoncharnes/pay/blob/master/lib/pay/billable.rb#L29

So more like:
`update_card(card_token) if card_token.present?`

This does mean the method is consistent in both cases. As long as you have the `card_token` set before this, it will either update the card for existing customers, or it will create a new customer with that card. I think this is correct for what we want here.

PR #11, opened 2017-11-03
Setting a card_token on a customer, should update the card when the c…

excid3:
…ustomer is retrieved

### DIFF
--- a/lib/pay/billable/stripe.rb
+++ b/lib/pay/billable/stripe.rb
@@ -20,7 +20,9 @@ module Pay
         token = ::Stripe::Token.retrieve(token)
 
         return if token.card.id == customer.default_source
-        save_stripe_card(token, customer)
+        result = save_stripe_card(token, customer)
+        self.card_token = nil
+        result
       end
 
       def stripe_subscription(subscription_id)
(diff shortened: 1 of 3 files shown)


excid3:
cc @mickeytgl

lib/pay/billable/stripe.rb
```diff
@@ -20,7 +20,9 @@ def update_stripe_card(token)
         token = ::Stripe::Token.retrieve(token)
 
         return if token.card.id == customer.default_source
-        save_stripe_card(token, customer)
+        result = save_stripe_card(token, customer)
+        self.card_token = nil
```
excid3: We want to reset the `card_token` attribute here after the token was updated on Stripe.

mickeytgl: Did this work out for you?? I tried to run this but again got an error. 
image

I tried to dive in to the code, but I'm getting pretty confused in the `save_stripe_card(token, customer)` and `update_stripe_card(token)` methods

excid3: I'm not sure what you mean?

PR #29, opened 2018-12-03
Change default_source_id->default_source on customer

jespr:
I'm pretty sure the Stripe::Customer object just has a `default_source` and not `default_source_id` per https://stripe.com/docs/sources/customers

It at least blows up in my app when trying to access `default_source_id` on `Stripe::Customer:

cc: @excid3


jasoncharnes:
This did change. I noticed this when working on tests, thank you for getting this in.

PR #31, opened 2018-12-16
Update subscription renewing/updated

jespr:
Updated subscription renewing/updated to match the new
processor/processor_id table layouts - as well as using `object.id` instead of `object.subscription.

cc: @excid3


excid3:
That was my mistake doing a copy-paste. :) Thanks for catching this!

PR #32, opened 2018-12-16
Chargeable class

jespr:
A couple of changes in this PR:
- make it possible to define your own `Charge` class (that means that I've namespaced the existing Charge-class -> Pay::Charge). Reason for this, is so you can define your own charge class with your own `receipt` and `filename` method.
- added a check in `Pay::Stripe::ChargeSucceeded` to check that the `charge` has a `receipt` method before attempting to mail a receipt email.

This fixes some bugs we're currently running in to, where `charge.receipt.render` errors in the mailer - as well as `charge.filename`

cc: @excid3

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,8 @@ end
 
 ## Emails
 
+### Stripe
+
 Emails can be enabled/disabled using the `send_emails` configuration option (enabled per default). When enabled, the following emails will be sent:
 
 - When a charge succeeded


excid3:
This brought a couple things to mind:

* I had considered adding receipts to this as a default feature. Thoughts? We'd want a way to override it, but maybe if `Receipts` gem was installed, it could include the `receipt` method?

* I like the idea of namespacing the gem's models regardless that way we don't interfere with anything in your app (like an email Subscription model or something). We probably don't need the table name customization since we can just move the tables to `pay_charges` and `pay_subscriptions`.

What if we modified `Pay::Charge` to be like:

```
class Pay::Charge < ApplicationRecord
  include Pay::Chargeable

  scope :sorted, ->{ order(created_at: :desc) }
  default_scope ->{ sorted }

  def filename
    "receipt-#{created_at.strftime("%Y-%m-%d")}.pdf"
  end
end
```

And the `Chargeable` model optionally includes something like:

```ruby

  def receipt
    Receipts::Receipt.new(
      id: id,
      product: Jumpstart.config.application_name,
      company: {
        name: Jumpstart.config.business_name,
        address: Jumpstart.config.business_address,
        email: Jumpstart.config.support_email,
        #logo: Rails.root.join("app/assets/images/logo.png")
      },
      line_items: line_items,
    )
  end

  def line_items
    line_items = [
      ["Date",           created_at.to_s],
      ["Account Billed", "#{owner.name} (#{owner.email})"],
      ["Product",        Jumpstart.config.application_name],
      ["Amount",         ActionController::Base.helpers.number_to_currency(amount / 100.0)],
      ["Charged to",     "#{card_type} (**** **** **** #{card_last4})"],
    ]
    line_items << ["Additional Info", owner.extra_billing_info] if owner.extra_billing_info?
    line_items
  end

**Comment by jespr** [comment]
I'm fine with that idea. I do think we should make another check, to see if the user has added the required information from Pay.config as well (name, address support email) - and only send out a receipt if they had. There's too much stuff that can go wrong otherwise, thoughts?

**Comment by jespr** [comment]
..or even break out the send_emails bool to three new ones: `send_charge_succeeded_email`, `send_charge_refunded_email` and `send_subscription_renewing_email`

File: app/models/concerns/pay/chargeable.rb [hunk 1]
```diff
@@ -34,5 +36,40 @@ def stripe_refund!(amount)
...
+
+      def line_items
+        line_items = [
+          ["Date",           created_at.to_s],
+          ["Account Billed", "#{owner.name} (#{owner.email})"],
+          ["Product",        Pay.config.application_name],
+          ["Amount",         ActionController::Base.helpers.number_to_currency(amount / 100.0)],
+          ["Charged to",     "#{card_type} (**** **** **** #{card_last4})"],
+        ]
+        line_items << ["Additional Info", owner.extra_billing_info] if owner.extra_billing_info?
```
excid3: `extra_billing_info` will be something we will need to add to the User model. This is important for adding things like VAT IDs and such.

excid3: We can add that field to this migration: https://github.com/jasoncharnes/pay/blob/master/db/migrate/20170503131610_add_fields_to_users.rb

app/models/concerns/pay/chargeable.rb
```diff
@@ -36,5 +36,40 @@ def stripe_refund!(amount)
...
+        )
+      end
+
+      def line_items
+        line_items = [
+          ["Date",           created_at.to_s],
+          ["Account Billed", "#{owner.name} (#{owner.email})"],
+          ["Product",        Pay.config.application_name],
+          ["Amount",         ActionController::Base.helpers.number_to_currency(amount / 100.0)],
+          ["Charged to",     "#{card_type} (**** **** **** #{card_last4})"],
```
excid3: This will also need an update in the future, since we'll have PayPal as well. `card_type` will be `PayPal` and then `card_last4` will be the email address of the account.

This is only for Braintree functionality. I'm going to push up some code for their webhooks that I wrote yesterday.

README.md
```diff
@@ -103,6 +105,13 @@ Pay.setup do |config|
 end
 ```
 
+## Emails
+
+Emails can be enabled/disabled using the `send_emails` configuration option (enabled per default). When enabled, the following emails will be sent:
+
+- When a charge succeeded
+- When a charge was refunded
```
excid3: Just realized, since Braintree doesn't have webhooks for charges, we'll need to probably send emails directly for those. 

https://developers.braintreepayments.com/guides/webhooks/overview

app/models/concerns/pay/chargeable.rb
jespr: @excid3 how do you update the two schema.rb's (test/dummy/db and test/support) ?

excid3: If I remember right, we just modify the schema.rb files? It's using an SQLite in-memory db for it, so I believe it just loads from the schema file.

app/models/concerns/pay/chargeable.rb
jespr: We can always extract this to a method that looks at the processor for the charge type and does it's thing - but that's for a later PR

jespr:
@excid3 might taking another look? I think I addressed your comments.

PR #35, opened 2018-12-24
Sync email changes to processor

jespr:
This should sync the local email changes to the payment processor if the user who's email got updated has any subscriptions.

@excid3 not sure if we want to do this in a background job instead?

### DIFF
--- a/lib/pay/billable.rb
+++ b/lib/pay/billable.rb
@@ -12,6 +12,8 @@ module Pay
       has_many :charges, foreign_key: :owner_id
       has_many :subscriptions, foreign_key: :owner_id
 
+      after_save :update_processor_email
+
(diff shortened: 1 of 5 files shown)


lib/pay/billable.rb
```diff
@@ -83,6 +85,12 @@ def upcoming_invoice
 
     private
 
+    def update_processor_email
+      if saved_change_to_email? && subscriptions.any?
```
excid3: I think we can swap the `subscriptions.any?` with `processor_id?` that way it updates them even if they don't have a subscription, but may have previously.

jespr:
Great idea! Been struggling a little with the lifetime of changes and what is/isn’t kept once you’re not subbed anymore :)

PR #36, opened 2018-12-24
Fix some small mailer bugs

jespr:
I discovered some more issues with the mailers, the subject field appeared to be named `From` - and it appears that the column name on charge is `card_type` and on users it's `card_brand` - we probably want to be consistent here?

### DIFF
--- a/app/mailers/pay/user_mailer.rb
+++ b/app/mailers/pay/user_mailer.rb
@@ -5,8 +5,8 @@ module Pay
 
       attachments[charge.filename] = charge.receipt.render
       mail(
-        to: "#{user.name} <#{user.email}>",
-        subject: "Payment receipt"
+        to: to(user),
+        subject: "Payment receipt",
       )
     end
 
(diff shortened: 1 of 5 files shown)


app/views/pay/user_mailer/receipt.html.erb
```diff
@@ -7,7 +7,7 @@ RECEIPT - SUBSCRIPTION<br/>
 <br/>
 Amount: USD <%= ActionController::Base.helpers.number_to_currency(@charge.amount / 100.0) %><br/>
 <br/>
-Charged to: <%= @charge.card_brand %> (**** **** **** <%= @charge.card_last4 %>)<br/>
+Charged to: <%= @charge.card_type %> (**** **** **** <%= @charge.card_last4 %>)<br/>
```
excid3: This reminded me, I want to go through and standardize everything to either `card_brand` or `card_type`. I believe Stripe uses `brand` and Braintree uses `type`. Right now we're generally processing credit cards + PayPal, but Braintree supports Venmo and some other options. I'm not sure the best wording for this, but I'll open up another issue for it.

excid3: For now I think we'd be fine to merge this, and then fix it in another commit.

jespr:
I discovered some other things that were missing when I attempted to write tests for the configurable subject - so I'm going to add these fixes to this PR as well :)

app/mailers/pay/user_mailer.rb
```diff
@@ -5,25 +5,35 @@ def receipt(user, charge)
...
-        subject: "Your upcoming subscription renewal",
+        to: to(user),
+        subject: Pay.email_renewing_subject,
       )
     end
+
+    private
+
+    def to(user)
+      if user.respond_to?(:name)
```
excid3: Great addition, thanks!

jespr:
@excid3 unless you have any other comments, this is ready to go

PR #39, opened 2018-12-28
Added email views generator

jespr:
Make it so you can copy the email views to your app if you want to customize them (https://github.com/jasoncharnes/pay/issues/34)


excid3:
Thanks @jespr 🥇

PR #41, opened 2018-12-30
Fix on generic trial

jespr:
I think the logic in `on_generic_trial?` was wrong. This PR fixes that

cc: @excid3


excid3:
LOL, yeah that was backwards. Thanks for catching that!

PR #42, opened 2018-12-30
Remove receipt_attributes check for now

jespr:
I introduced a pretty nasty bug when I tried to check for `Pay.config` attributes. I don't think I know ruby well enough to fix it - so let's just stop being this defensive for now and remove those checks.

cc: @excid3


excid3:
I refactored this a bit, so I wasn't sure if this would be working properly or not.

I'm fine with taking this out. Also I think your test including the module won't be super helpful unless we have the method for testing the required attributes, so I'm going to just leave that out as well.

PR #46, opened 2019-01-03
More webhook fixes

jespr:
This updates to use the exception throwing bang methods.

I have also updated the renewing class, we get passed an invoice-object so we need to use `event.data.subscription` to find the subscription instead of `object.id`

cc: @excid3


jespr:
I have more tests coming, these stripe ones are just such a slog

PR #50, opened 2019-01-08
Added more Webhook tests

jespr:
Added some more tests of stripe webhook functionality.

Also fixed a bug in `subscription_updated` where `subscription.processor_id  = object.plan.id` should be `subscription.processor_plan  = object.plan.id`

cc: @excid3


excid3:
Awesome, thanks so much!

PR #53, opened 2019-01-21
Update README.md

jean-francois-labbe:


excid3:
Huge thanks @jean-francois-labbe 🤘

PR #56, opened 2019-01-21
configure travis to test with ruby 2.4.1, 2.5.3 and 2.6.0

jean-francois-labbe:
What do you think of testing against ruby 2.4.1, 2.5.3 and 2.6.0?


excid3:
Works for me! I don't think we have any version specific code in our gem, but the dependencies might, so it may catch something.

PR #58, opened 2019-01-25
Use Configured billable model instead of User. in webhooks

jean-francois-labbe:
There was an issue when using webhooks and a `billable_class` different than `User`.


excid3:
Those have been on my todo list to fix. Thanks!

PR #71, opened 2019-03-03
Added `dependent: destroy` to Billable :charges :subscriptions

eelcoj:
So all charges + subscriptions will be removed if Billable/User gets destroyed.

Not 100% certain—thus draft PR—if this has any side-effects, but otherwise this is behaviour one wants.

### DIFF
--- a/lib/pay/billable.rb
+++ b/lib/pay/billable.rb
@@ -7,8 +7,8 @@ module Pay
     included do
       include Pay::Billable::SyncEmail
 
-      has_many :charges, class_name: Pay.chargeable_class, foreign_key: :owner_id, inverse_of: :owner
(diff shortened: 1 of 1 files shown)


excid3:
This is an interesting one. On one hand, we might want to remove those records. On the other hand, we may want to keep them for historical purposes. 

If you remove charges, you lose historical data about purchases. Of course, you have them in Stripe or Braintree, but it feels like we should preserve that data here too.

eelcoj:
My thinking here is: if User wants their data removed, you should remove their data (GDPR + general good practice). 
As such I might even add a call to Stripe to remove from there too, but as an optional argument.

Could make it optional altogether?

eelcoj:
Just leave this for now.

PR #74, opened 2019-03-11
give env variables higher precedence

jean-francois-labbe:


excid3:
You're too fast. 😜

jean-francois-labbe:
Thanks and I really need it :smile:

excid3:
Merged! I'll cut a new release real fast as well.

PR #77, opened 2019-06-04
Allow mount path to be customised

JoeSouthan:
:wave:

I found that I needed to control where Pay::Engine mounts in my own project as I have a catch-all route. Pay::Engine currently appends to the route list which places it after the catch-all.

I've kept the current functionality by adding 2 configuration options:
- `automount_webhook_routes` for turning off the automatic mounting
- `webhooks_path` for customising where the engine mounts automatically

I also came across a few issues while trying to introduce an integration test for the routing.
- BraintreeController had an undefined method
- In test_helper.rb, the ordering of the gem requirements needed a bit of tweaking. Bundler was not requiring the payment provider gems when the dummy app boots as they are listed as development dependencies.

Let me know what you think!


excid3:
Thanks @JoeSouthan! I'll try and take a look shortly!

jasoncharnes:
@JoeSouthan I'm going to let @excid3 give it the official okay when he has time to look at it, but I wanted to chime in and say that:

1. It looks good to me. 👍 
2. Thank you for your contribution! 🎉

excid3:
Thanks @JoeSouthan 🤘 

Just released this as 1.0.0!

PR #83, opened 2019-08-17
Provide method that checks for active trial or subscription

jespr:
Added a convenience method to billable that checks for either an active trial, or a subscription


excid3:
You're the man! 🤘

jasoncharnes:
Thanks @jespr, I needed this the other day.

PR #94, opened 2019-11-30
Get rid of ambiguous column with team based scopes.

caseyprovost:
When trying to add scopes to the `Team` class in jumpstart (but applicable in other scenarios) we run into ambiguous column error. This should remove that.

### DIFF
--- a/app/models/pay/subscription.rb
+++ b/app/models/pay/subscription.rb
@@ -14,7 +14,7 @@ module Pay
 
     # Scopes
     scope :for_name, ->(name) { where(name: name) }
-    scope :on_trial, ->{ where.not(trial_ends_at: nil).where("? < trial_ends_at", Time.zone.now) }
+    scope :on_trial, ->{ where.not(trial_ends_at: nil).where("? < #{table_name}.trial_ends_at", Time.zone.now) }
     scope :cancelled, ->{ where.not(ends_at: nil) }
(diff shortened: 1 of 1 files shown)


app/models/pay/subscription.rb
```diff
@@ -14,7 +14,7 @@ class Subscription < ApplicationRecord
 
     # Scopes
     scope :for_name, ->(name) { where(name: name) }
-    scope :on_trial, ->{ where.not(trial_ends_at: nil).where("? < trial_ends_at", Time.zone.now) }
+    scope :on_trial, ->{ where.not(trial_ends_at: nil).where("? < #{table_name}.trial_ends_at", Time.zone.now) }
     scope :cancelled, ->{ where.not(ends_at: nil) }
     scope :on_grace_period, ->{ cancelled.where("? < ends_at", Time.zone.now) }
```
excid3: Shouldn't we add it here too?

excid3:
Looks great, thanks! 👍

app/models/pay/subscription.rb
caseyprovost: Makes sense to me!

rikkipitt:
Hi folks, I'm running into an issue where this scope specificity could help:

E.g.

`User.joins(:pay_subscription).merge(Pay::Subscription.active)`

However, master doesn't seem to include this merge. Is it outdated?

rikkipitt:
Ah, this was lost in 89c590c

PR #97, opened 2020-01-03
Only queue the sync if the change is committed

lostmahbles:
Otherwise you could queue up a job if the record isn't saved/created. This causes failures in Rails 6+.


excid3:
Thanks @lostmahbles! Is this in the case that an `update` failed with a rollback, it would still do the sync even though it shouldn't?

lostmahbles:
Exactly.

excid3:
Awesome, I'll apply this to the new `sca` branch as well.

PR #116, opened 2020-02-11
Update pay confirmation page

rikkipitt:
This prevents the payment intent confirmation page from "jumping" around as the Stripe modal appears/closes.

I believe the idea was to place the panel vertically central on the page, this pull request fixes it.

Alternatively, the main `application.html.erb` could be tweaked as it also has a few `h-full` css classes.

Rikki

### DIFF
--- a/app/views/layouts/pay/application.html.erb
+++ b/app/views/layouts/pay/application.html.erb
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
-<html>
-<head class="h-full">
+<html class="h-full">
+<head>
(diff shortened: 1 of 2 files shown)


rikkipitt:
Screenshot 2020-02-11 at 23 27 11

Versus

Screenshot 2020-02-11 at 23 26 17

rikkipitt:
I also noticed a minor mistake in the `application.html.erb` which I fixed.

excid3:
Wonderful. Thanks for fixing those!

PR #118, opened 2020-02-23
Handle no `credentials` on Rails < 5.2

digitalmoksha:
If running on Rails < 5.2, then `credentials` does not exist on `Rails.application`.  Added code to tolerate this.


excid3:
Oh that's a great point. I've only been using this with Rails 5.2+ so I forgot that we'd need this. 🤘

PR #120, opened 2020-02-23
Allow running migration on Rails 4.2

digitalmoksha:
I'm currently integrating into a Rails 5.0 app.  The README mentions that Rails 4.2 and above is supported.

However running the migrations gives `ArgumentError: Unknown migration version "5.1"; expected one of "4.2", "5.0"` since two fo the migrations are targeted to `5.1` and `5.2`

I didn't notice anything 5.x specific in the migrations, so changed their version number to support 4.2


excid3:
Thanks for this! 👍

PR #122, opened 2020-02-29
Sets status back to active when resuming a subscription

jasoncharnes:
When resuming a subscription that's on a grace period, the status stays `canceled`.

This will set the status back to active when resuming.


excid3:
Since webhooks are required, this is a little redundant but it will help with displaying the right status immediately.

Which reminds me, I think on cancel-at-end-of-period, we set it to `canceled` too soon and probably need to fix that.

PR #126, opened 2020-03-04
Fix subscription status scopes

rikkipitt:
This adds the scope specificity back in that was introduced in
2f61c38f7c8a55a03d1a3b7ea0f1e4d504a4d73f but lost in 89c590cbbd21c2120ff35b60a75410a10e268fdc.

Originally committed by: @caseyprovost

Related: #94


excid3:
Thanks @rikkipitt 👍

PR #141, opened 2020-04-09
Updating docs to fill gaps for AddFieldsToBillable migration

### LINKED ISSUE #140: Documentation is slightly outdated?
There seems to be a mismatch between documentation and actual functionality. Initial `bin/rails pay:install:migrations` command created 4 migrations, but documentation suggests that it should be 3.

Adding extra 'AddFieldsToBillable.rb' to the list of new migrations.

Content of this migration is really interesting, it creates additional columns for billable table and how to change it. 

Seems like documentation is slightly outdated. But it's a really important part of documentation as it seems.

skatkov:
I'm currently going through the installation process and trying to update documentation according to actual behavior. 

Hopefully, this fixes #140

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,9 @@ To add the migrations to your application, run the following migration:
 
 `$ bin/rails pay:install:migrations`
 
-This will install four migrations. But before running these migrations, make sure to pick one of your models as a billable model. By default we're using `User` model as a Billable model. In case you want to change that, add couple of lines to pay initializer file `config/initializers/pay.rb`.
+This will install four migrations. 
+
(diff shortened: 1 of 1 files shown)


excid3:
This may be a great time to question if we need the readme to include these specifics? Maybe we should just remove these notes since the migrations are likely to change in the future.

skatkov:
@excid3 great suggestion. I removed list of migration files.

oleingemann:
Please continue your work @skatkov , the current docs are terrible. Lost 2 hours now on misc errors, not even getting through rake db:migrate 🥴

skatkov:
@oleingemann would love it if you contribute too -- I'm in a middle of a process myself :P

skatkov:
I think I'm done with this. Can we merge @excid3 ?

excid3:
Sorry it took me so long. Had some family stuff going on unfortunately. ☹️

Issue #148, opened 2020-04-26
Mailer - Undefined method `name' for User


dankmitchell:
The mailers are attempting to find [name](https://github.com/pay-rails/pay/blob/master/app/mailers/pay/user_mailer.rb#L47) from the user object but as far as I can tell that attribute doesn't exist. 

Should it be `customer_name` - https://github.com/pay-rails/pay/blob/master/lib/pay/billable.rb#L45

excid3:
Billable expects a `name` attribute on the model. If yours is called `customer_name` just delegate to that.

```ruby
def name
  customer_name
end
```

dankmitchell:
I have added a name method to by billable model to solve this but definitely can't see `name` anywhere in Pay... will close.

excid3:
Whoops, I meant first_name, last_name. I typically use Basecamp name_of_person gem for billing and that's why were' using those columns.

dankmitchell:
Yes, exactly. The billable model already has `customer_name` so I would change the mailer from `user` to `customer_name`. Happy to raise a PR for this but I wanted to check I wasn't going mad.

excid3:
Good catch. I think my intention was to use `customer_name` everywhere so it wasn't likely to conflict with anything on your model and it could be easily overridden. Happy to have a PR for that.

PR #149, opened 2020-04-27
Udpate name to customer_name in useer mailer

dankmitchell:
Fixes https://github.com/pay-rails/pay/issues/148

The user mailer was incorrectly looking for the `name` attribute but the [billable](https://github.com/pay-rails/pay/blob/master/lib/pay/billable.rb#L45) model only has a `customer_name` attribute.


excid3:
Thanks @dankmitchell 🤘

PR #162, opened 2020-06-02
Subscriptions: Billable model with string as ID will not be saved properly

skatkov:
# Problem
I have a Billable object called store that uses string as an id:
```
create_table "stores", id: :string, force: :cascade do |t|
    ...
    t.index ["id"], name: "index_stores_on_id", unique: true
  end
```

In my particular case, after user subscribes -- owner_id ends up missing, because of type mismatch..

# Proposed solution
If we change owner_id to string type, it will work in case of ID being a string and integer. I think it's a safer default migration.

### DIFF
--- a/db/migrate/20170205020145_create_pay_subscriptions.rb
+++ b/db/migrate/20170205020145_create_pay_subscriptions.rb
@@ -1,7 +1,7 @@
 class CreatePaySubscriptions < ActiveRecord::Migration[4.2]
   def change
     create_table :pay_subscriptions do |t|
-      t.references :owner
+      t.references :owner, type: :string
       t.string :name, null: false
       t.string :processor, null: false
       t.string :processor_id, null: false


excid3:
Hmm, I'm on the fence with this. I can see it being a gotcha if you're doing nonstandard IDs, but I'd also prefer to stick with the Rails defaults.

skatkov:
@excid3 maybe we should add a note into readme about this gotcha?

I didn't test this thoroughly, but my understanding is that string column should work for both cases (default "integer" and "string")?

Another possible solution is trying to atch the error and notify users to change column type.

excid3:
I think a note in the README would be fine. Might also want to leave a comment in the migration.

It's not standard to have a string ID column, so I think that it's probably fair to have a note in the docs rather than changing the code for this.

skatkov:
Closing this one, but #167 should add a required note to database migration.

PR #183, opened 2020-07-10
Marketplace Webhooks

### LINKED ISSUE #178: Webhook Management for Connected Accounts (Marketplace)
Hi all, thanks for your work on this gem, super useful!

I'm building an application that charges customers on behalf of a connected business.

Creating marketplace charges or payment methods through the Stripe API works well, but when the webhooks fire they fail with errors like: `Pay::Error (No such customer: customer_id)`.

Presumably this is happening because webhook events are attempting to retrieve a customer from the platform account, instead of the connected business.

I'm aware from reading other issues that marketplace support is on your radar, but wanted to provide a thread for ideas and possible implementation work.

One idea is to leverage the StripeEvent gem and expose a second webhooks endpoint for connected accounts.

Any insight or ideas appreciated.

Screen Shot 2020-06-30 at 4 59 52 PM

mrjonesbot:
@excid3 Did a first pass here, attempting to run tests but the suite needs Stripe keys, where are storing those locally?

Resolves #178

### DIFF
--- a/lib/pay/billable.rb
+++ b/lib/pay/billable.rb
@@ -24,6 +24,7 @@ module Pay
       attribute :plan, :string
       attribute :quantity, :integer
       attribute :card_token, :string
+      attribute :marketplace_id, :string
       attribute :pay_fake_processor_allowed, :boolean, default: false
 
       validate :pay_fake_processor_is_allowed, if: :processor_changed?
--- a/lib/pay/stripe/billable.rb
+++ b/lib/pay/stripe/billable.rb
@@ -27,7 +27,7 @@ module Pay
       # Returns Stripe::Customer
       def customer
         stripe_customer = if processor_id?
-          ::Stripe::Customer.retrieve(processor_id)
+          retrieve_customer_with_marketplace
         else
           sc = ::Stripe::Customer.create(email: email, name: customer_name)
           billable.update(processor: :stripe, processor_id: sc.id)
@@ -44,6 +44,12 @@ module Pay
         stripe_customer
       rescue ::Stripe::StripeError => e
         raise Pay::Stripe::Error, e
+      end
+
+      def retrieve_customer_with_marketplace
+        ::Stripe::Customer.retrieve(processor_id, {
+          stripe_account: marketplace_id
+        })
       end
 
       # Handles Billable#charge
@@ -148,6 +154,32 @@ module Pay
 
       def create_setup_intent
         ::Stripe::SetupIntent.create(customer: processor_id, usage: :off_session)
+      end
+
+      def payment_method_with_marketplace(payment_method_id)
+        ::Stripe::PaymentMethod.retrieve(
+          payment_method_id,
+          { stripe_account: marketplace_id }
+        )
+      end
+
+      private
+
+      def create_stripe_customer
+        customer = ::Stripe::Customer.create(email: email, name: customer_name)
+        update(processor: "stripe", processor_id: customer.id)
+
+        # Update the user's card on file if a token was passed in
+        if card_token.present?
+          payment_method = ::Stripe::PaymentMethod.attach(card_token, {customer: customer.id})
+          customer.invoice_settings.default_payment_method = payment_method.id
+          customer.save
+
+          update_stripe_card_on_file ::Stripe::PaymentMethod.retrieve(card_token).card
+        end
+
+        customer
+>>>>>>> 356f27f (Stripe Connected Customer & Card Sync)
       end
 
       def trial_end_date(stripe_sub)
--- a/lib/generators/active_record/pay_generator.rb
+++ b/lib/generators/active_record/pay_generator.rb
@@ -40,6 +40,7 @@ module ActiveRecord
       t.string :card_last4
       t.string :card_exp_month
       t.string :card_exp_year
+      t.string :marketplace_id
       t.text :extra_billing_info
 RUBY
       end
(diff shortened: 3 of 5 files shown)


lib/pay/stripe/webhooks/customer_updated.rb
```diff
@@ -4,12 +4,14 @@ module Webhooks
...
           object = event.data.object
+          account_id = event.account
+
           billable = Pay.find_billable(processor: :stripe, processor_id: object.id)
 
           # Couldn't find user, we can skip
           return unless billable.present?
 
-          billable.sync_card_from_stripe
+          billable.sync_card_from_stripe(account: account_id)
```
mrjonesbot, line 14: Instead of passing in `account` to `sync_card_from_stripe`, maybe we can set the value on `billable` with an attr_accessor.

excid3:
@mrjonesbot I'm just doing `STRIPE_PRIVATE_KEY=x rails test`

lib/pay/stripe/webhooks/customer_updated.rb
excid3: If the account_id was already stored on the Billable, wouldn't we already be passing in the `account_id` to `Pay.find_billable`?

Then we'd have it on the model already and wouldn't need to pass it in here.

mrjonesbot: Right -- I've adjusted the migration fields to include `marketplace_id`, then checks if there's a stored value on billable when retrieving the customer.

The other piece of the puzzle is automagically capturing the connected account's id on `marketplace_id`, looking through event types that correspond to creating a connected account.

mrjonesbot: It looks like no hooks are triggered when a marketplace business connects with Stripe, so storing the `marketplace_id` must be handled in the oauth redirect callback.

lib/pay/stripe/billable.rb
```diff
@@ -6,14 +6,24 @@ module Billable
...
       end
 
+      def retrieve_customer_with_marketplace
+        if marketplace_id?
+          ::Stripe::Customer.retrieve(processor_id, {
+            stripe_account: marketplace_id
+          })
+        else
+          ::Stripe::Customer.retrieve(processor_id)
+        end
```
GeorgeRDBuckley, line 24:
```suggestion
          ::Stripe::Customer.retrieve(processor_id, {
            stripe_account: marketplace_id
          })
```

I think that if `marketplace_id` is `nil`, this will work exactly the same, so the `if/else` is unnecessary.

excid3:
@mrjonesbot I added a ENV var you can use to disable VCR for testing. https://github.com/excid3/pay/blob/master/test/test_helper.rb#L37

Looks like I never pushed that up to pay-rails/pay, so I'm going to do that now.

lib/pay/stripe/billable.rb
excid3: If that works as normal when passing `nil`, that would be fantastic and would simplify things a lot.

GeorgeRDBuckley: https://stripe.com/docs/api/customers/retrieve

I can't see it anywhere in the docs, but it appears that `stripe_account` is an optional param on retrieve, so passing `nil` would work exactly the same.

I have also manually tested this on a demo connect app I have and passing `nil` retrieves a customer from the main stripe account, passing an `account_id` retrieves the customer from the connected account.

excid3: Great, sounds good then. I know sometimes API gems like this will validate an argument like that, but as usual, the Stripe gem is really, really well built.

mrjonesbot: Sweet, I'll adjust and test. I kind of made the same assumption about passing nil arguments.

mrjonesbot: Thanks for the suggestion, I made the same assumption about passing `nil`.

excid3:
@mrjonesbot what API version is your Stripe account set to? Maybe it's pre-SCA?

excid3:
I've been using a late 2018 API version, so possibly this change is one that's affecting it: https://stripe.com/docs/upgrades#2019-03-14

excid3:
Well, we should first confirm if that's the case. If so, then we might do that, or we might just require everyone's on a newer version of the API.

I actually looked at the wrong Stripe account, so I'm not sure my Pay test account is actually that old of an API version.

excid3:
Awesome! I'm just about to leave for vacation, but I'll try and review this. Not sure why that one test is failing with no API key being set. That should be done through the github actions secrets. It looks like this is turning out to be a lot simpler than I thought it would be!

lib/pay/billable.rb
```diff
@@ -56,10 +57,10 @@ def subscribe(name: "default", plan: "default", **options)
       send("create_#{processor}_subscription", name, plan, options)
     end
 
-    def update_card(token)
+    def update_card(token, marketplace: true)
```
mrjonesbot, line 60: @excid3 As indicated in my latest comment, once a Pay model stores the `marketplace_id`, then all interactions with Stripe are scoped to that connected account.

However some interactions, like updating the account card to pay the platform, break as they are no longer able to interact with their platform card/customer.

Feeding the `marketplace` flag through down to `stripe_customer` was the only obvious way I could see to ensure that I could update the platform card + subscriptions while still enabling marketplace interactions in other areas of my application.

This change is more for my short term benefit but we should talk about what the API should look like ideally -- open to suggestions.

PR #190, opened 2020-08-10
potential error in Brainstree#set_up

kieranklaassen:
does this need to be `Pay::Braintree::Billable` instead?


excid3:
:O Wow, surprised that I missed that. Not even sure how the Braintree stuff would work without that...

PR #191, opened 2020-08-11
Add idempotency key as a second hash

drewscott:


lib/pay/stripe/billable.rb
```diff
@@ -47,7 +52,7 @@ def create_stripe_charge(amount, options = {})
       # Handles Billable#subscribe
       #
       # Returns Pay::Subscription
-      def create_stripe_subscription(name, plan, options = {})
+      def create_stripe_subscription(name, plan, options = {}, idempotency_key = SecureRandom.uuid)
```
excid3, line 55: What if we have the idempotency key as part of options and remove it when passed in like we do with quantity? That way it doesn't change the method signature and we can seamlessly remove it like we do with the `quantity`

excid3:
Let's also add a test where we try and make a charge with the same idempotency key and make sure the second attempt fails with the correct error.

excid3:
Closing this because of https://github.com/pay-rails/pay/commit/d7c929190a3a7667f3db36da759d7feec5aac3e9

PR #194, opened 2020-09-01
Allow pay 'Go back' button to be configured via ?back=/path/to/item

drnic:
Currently the "Go back" button is linked to `root_path`. This PR allows a `?back=/path/to/item` attribute to be provided to the `/payments/pi_xyz?back=%2Fpath` URL

### DIFF
--- a/app/controllers/pay/payments_controller.rb
+++ b/app/controllers/pay/payments_controller.rb
@@ -1,6 +1,7 @@
 module Pay
   class PaymentsController < ApplicationController
     def show
+      @redirect_to = params[:back].presence || root_path
       @payment = Payment.from_id(params[:id])
     end
   end
--- a/app/views/pay/payments/show.html.erb
+++ b/app/views/pay/payments/show.html.erb
@@ -51,7 +51,7 @@
         </div>
       <% end %>
 
-      <%= link_to t("back"), root_path, class: "inline-block w-full px-4 py-3 bg-gray-200 hover:bg-gray-300 text-center text-gray-700 rounded-lg" %>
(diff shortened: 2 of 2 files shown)


app/controllers/pay/payments_controller.rb
```diff
@@ -1,6 +1,7 @@
 module Pay
   class PaymentsController < ApplicationController
     def show
+      @redirect_to = params[:back] || root_path
```
excid3, line 4: We should probably change this to `params[:back].presence || root_path` so that an empty string is handled as nil.

drnic: Ahh very nice. Wasn't aware of `.presence`

excid3:
Released as 2.2.1 👍

drnic:
I'm not seeing this PR in v2.2.1 -- am I failing at `bundle update` or is it possible the new gem doesn't includes this PR?

image

excid3:
Hmm! Maybe I goofed and didn't get the right thing tagged. Will check this afternoon and re-release or something.

asecondwill:
Is this added in yet? I'm adding like this

gem 'pay', '~> 2.2.1'

Then in my Subscription create action:

rescue Pay::ActionRequired => e
    redirect_to pay.payment_path(e.payment.id, back: '/retailer/dash')

But it is still redirecting to root. 

does not seem to be updated:
https://github.com/pay-rails/pay/blob/1d7a024b91c1f5f13a8b72bdd33af13374b694f9/app/views/pay/payments/show.html.erb#L54

excid3:
Doesn't look like this was ever released. I'll cut 2.2.2 right now for it.

asecondwill:
thanks!

I've noticed the generator has put that file in my views dir, so I can keep going by just updating locally.

excid3:
And released!

PR #208, opened 2020-10-06
Fix polymorphic querying for charges and subscriptions

### LINKED ISSUE #207: Subscription <-> Billable relation is polymorphic in one direction only
It looks like this line in billable.rb:
```ruby
has_many :subscriptions, class_name: Pay.subscription_class, foreign_key: :owner_id, inverse_of: :owner
```
is missing either `as` or `foreign_type` to correctly take into account the polymorphism when going from billable to subscription.

Here's an example that illustrates the problem:
```
[71] pry(main)> Pay::Subscription.last.owner
  Pay::Subscription Load (5.8ms)  SELECT "pay_subscriptions".* FROM "pay_subscriptions" ORDER BY "pay_subscriptions"."id" DESC LIMIT $1 /*line:(pry):66:in `__pry__'*/  [["LIMIT", 1]]
  User Load (6.2ms)  SELECT "users".* FROM "users" WHERE "users"."id" = $1 LIMIT $2 /*line:(pry):66:in `__pry__'*/  [["id", 2], ["LIMIT", 1]]
=> #<User id: 2, ... # <- correctly infers User from owner_type
[72] pry(main)> User.last.subscription
  User Load (5.0ms)  SELECT "users".* FROM "users" ORDER BY "users"."id" DESC LIMIT $1 /*line:(pry):67:in `__pry__'*/  [["LIMIT", 1]]
  Pay::Subscription Load (1.5ms)  SELECT "pay_subscriptions".* FROM "pay_subscriptions" WHERE "pay_subscriptions"."owner_id" = $1 AND "pay_subscriptions"."name" = $2 ORDER BY "pay_subscriptions"."id" DESC LIMIT $3 /*line:(pry):67:in `__pry__'*/  [["owner_id", 2], ["name", "default"], ["LIMIT", 1]]
=> #<Pay::Subscription:0x00007f96a3ef15c8
 id: 3 ... # <- here the query to load subscription only filters by owner_id, does not filter by owner_type
```

grk:
Before the change, `as: :owner` was missing from the relations from a billable to charges and subscriptions defined when including Billable into a model.

When two billables from different models had the same id, querying from one of them could return the subscriptions or charges from the other.

Fixes #207

### DIFF
--- a/lib/pay/billable.rb
+++ b/lib/pay/billable.rb
@@ -20,8 +20,8 @@ module Pay
       include Pay::Stripe::Billable if defined? ::Stripe
       include Pay::Braintree::Billable if defined? ::Braintree
 
-      has_many :charges, class_name: Pay.chargeable_class, foreign_key: :owner_id, inverse_of: :owner
-      has_many :subscriptions, class_name: Pay.subscription_class, foreign_key: :owner_id, inverse_of: :owner
+      has_many :charges, class_name: Pay.chargeable_class, foreign_key: :owner_id, inverse_of: :owner, as: :owner
+      has_many :subscriptions, class_name: Pay.subscription_class, foreign_key: :owner_id, inverse_of: :owner, as: :owner
 
       attribute :plan, :string
       attribute :quantity, :integer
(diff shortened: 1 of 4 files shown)


excid3:
Wonderful, thanks @grk !

PR #210, opened 2020-10-07
docs: add event types for stripe webhook

kieranklaassen:
I did not find this in the documentation so checked the source code and added to readme


excid3:
Thanks! 👍  I always just send all events, so it didn't cross my mind.

PR #213, opened 2020-10-21
Fix typo in error message

dankmitchell:
`beacuse` > `because`


excid3:
Good catch, thanks @dankmitchell 🙏

PR #238, opened 2020-12-14
Refactor emails

excid3:


archonic:
Is there a way to turn certain emails on/off or the turn off the Pay mailer altogether? Stripe has built out their email functionality and it looks like it's possible to configure a layout and let them handle all customer billing emails.

excid3:
Not yet, but that's a good feature to add here.

excid3:
@archonic I should say, you can turn off Pay emails in the config, but not specific ones atm. https://github.com/pay-rails/pay#configuration

archonic:
Ah, thank you. I think a master switch might be all I need. I'll find out in the coming weeks.

PR #239, opened 2020-12-18
No need to query for subscriptions again if eager loaded

donnguyen:
Say if we have subscriptions eager loaded by Billable model like below:
`Billable.includes(:subscriptions)`
Then we shouldn't use `.for_now` scope when we call `.subscription` on a Billable instance as this causes another DB query. Instead of that, we could find by name on the loaded association.

Sometimes (at least for my app) apps rely too much on `.subscription`, this helps reduce queries to DB quite a lot.


excid3:
Thanks @donnguyen 🙏

PR #240, opened 2020-12-18
Add paused_from column to subscriptions

nm:
Adds a `paused_from` column to pay subscriptions. 
With that new column it is much easier to determine if a subscription is or will be paused (similar to cancel).
`subscription.on_grace_period?` applies for cancellations and pauses.

### DIFF
--- a/db/migrate/20201217084605_add_paused_from_to_pay_subscriptions.rb
+++ b/db/migrate/20201217084605_add_paused_from_to_pay_subscriptions.rb
@@ -1,5 +1,0 @@
-class AddPausedFromToPaySubscriptions < ActiveRecord::Migration[4.2]
-  def change
-    add_column :pay_subscriptions, :paused_from, :datetime
-  end
-end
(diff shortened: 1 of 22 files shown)


db/migrate/20201217084605_add_paused_from_to_pay_subscriptions.rb
```diff
@@ -0,0 +1,5 @@
+class AddPausedFromToPaySubscriptions < ActiveRecord::Migration[4.2]
+  def change
+    add_column :pay_subscriptions, :paused_from, :datetime
```
excid3, line 3: Since Paddle is the only one that needs a `paused_from` column (at the moment), what do you think about just putting this in the json `data` column instead?

nm: If it makes merging this PR easier, it's fine for me to put this in the `data` column for now. It's always possible to do a refactoring later if an additional payment provider needs a pause functionality.

lib/pay/paddle/subscription.rb
```diff
@@ -8,6 +8,7 @@ module Subscription
 
         store_accessor :data, :paddle_update_url
         store_accessor :data, :paddle_cancel_url
+        store_accessor :data, :paddle_paused_from
```
excid3, line 11: Thanks for the change. I was just thinking that in v3 when I reorganize things, we can delegate to the a Paddle::Subscription class and let it handle pausing and that way the logic doesn't have to bleed into the other providers. 👍

PR #244, opened 2020-12-22
Translations

### LINKED ISSUE #228: How can I translate error messages?
Any chance for i18n support?

GeorgeRDBuckley:
* Adds translations to errors.
* Add a few more error types

closes #228
closes #229 also


GeorgeRDBuckley:
Braintree does Error's in a weird way, and I can't think of a sensible way of translating them.

lib/pay/braintree/billable.rb
```diff
@@ -31,9 +31,9 @@ def braintree_customer
           result.customer
         end
       rescue ::Braintree::AuthorizationError
-        raise BraintreeAuthorizationError
+        raise Pay::Braintree::AuthorizationError
```
excid3, line 34: This would be a breaking change. We would have to raise the old error with the deprecation warning until the next major release where we could replace this.

I very much like the new Error namespacing, so I definitely want to do that. If we `raise` the old error classes, we can replace this in the next major version bump.

lib/pay.rb
```diff
@@ -103,43 +103,74 @@ def self.receipts_supported?
...
+  class InvalidPaymentMethod < PaymentError
+    def message
+      I18n.t("errors.invalid_payment")
+    end
+  end
 
-    def initialize(payment)
-      @payment = payment
+  module Braintree
+    class Error < Error
```
excid3, line 134: We should have done this from the beginning. 😅

lib/pay/braintree/billable.rb
GeorgeRDBuckley: Yeah makes sense, I may have come to a similar thinking over the christmas break. So this was a good little reminder. 👍

GeorgeRDBuckley:
@excid3 This is ready for another review, I have pulled the error classes out into a seperate file (`/pay/errors.rb`) - so that we don't muddy up `pay.rb` when more payment processors are added.

excid3:
@GeorgeRDBuckley Awesome. Also I think we can extract the payment provider errors to go inside the payment providers modules so that they can be organized with the rest of the payment provider code. That would help organize things a bit more.

config/locales/en.yml
```diff
@@ -15,6 +15,132 @@ en:
     success: The payment was successful.
   all_rights_reserved: All rights reserved.
   back: Go back
+  receipt:
```
excid3, line 18: I just got realizing, we probably should namespace all of the translations under `pay` shouldn't we? That way we don't conflict with user specified translations for something simple like `back` that they may use in other places.

No need to do that in this PR, but just a thought I had.

PR #248, opened 2021-01-02
Configurable default names - Product/Plan

### LINKED ISSUE #220: Make subscription default name settable
I'm blowin up your inbox tonight but I think this is a good idea 😄 

I imagine most people would have a single subscription per billable model. Making the name optional makes sense but it requires a name of "default".
https://github.com/pay-rails/pay/blob/3a0cdfe4f74d28add7b02ff46fe04d0ac09f4e60/lib/pay/billable.rb#L92

It would be nice if we could set `config.default_product_name` or something like that in the pay initializer. Happy to make a PR if you agree.

GeorgeRDBuckley:
Adds options to configure the default names for product/plan.

closes #220

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,9 @@ Pay.setup do |config|
 
   config.send_emails = true
 
+  config.default_product_name = "default"
+  config.default_plan_name = "default"
+
   config.automount_routes = true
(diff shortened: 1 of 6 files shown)


excid3:
Fantastic, thanks @GeorgeRDBuckley 🙏

PR #251, opened 2021-01-08
Update subscription.rb

nm:
To conform with the error translations


excid3:
Thanks for all your help @nm 🙏

PR #254, opened 2021-01-13
Refactor subscription to delegate most logic to payment provider

### LINKED ISSUE #253: NoMethodError when resuming Stripe subscriptions
It seems that the `paddle?` check introduced [here](https://github.com/pay-rails/pay/blob/59ab4da98b89c5266f7cc4a9051a408a414913f6/app/models/pay/subscription.rb#L74) creates a NoMethodError when trying to resume Stripe subscriptions. 

If I understand the lib correctly, that symbol only exists for paddle subscriptions. When trying to resume, it's first checked if it's on its grace period and then if it's paused. When we hit the `paused?` method there's a check if it's a paddle sub, and the error follows.

excid3:
Should fix #253

* Moves payment provider helpers and scopes into Pay::Subscription so they're always accessible
* Pushes some pause & resume logic into the payment provider so they can control how it works


excid3:
@dzacarias give this a try and see if it fixes the issue?

dzacarias:
Yeah, it does!

PR #284, opened 2021-02-25
Fix no method error when cancelling trialing subscription

archonic:
I found this bug in my own test suite. It was just a missed delegation. This should also fix bugs in resuming and swapping trialing subscriptions since they also reference `trial_ends_at`.


excid3:
Awesome, thanks!

excid3:
I made all 3 payment providers subscription class delegates the same. Paddle had a similar issue.

Going to refactor some more and pull out a base class for each these so they can be shared and less likely to miss a delegate.

Issue #188, opened 2020-08-05
Prevent network issues using idempotent requests ?


drewscott:
This could help duplicated subscriptions and charges in the event of network issues. 

https://stripe.com/docs/api/idempotent_requests

excid3:
Yep! Since you need to pass in the same key each time, it probably needs to be done outside of Pay and passed in when you make a charge or subscription.

This gets sent in the second hash of options in the Stripe Ruby methods if someone wants to make a PR to pull out this option in the `charge` and `subscribe` args.

```ruby
charge = Stripe::Charge.create({
  amount: 2000,
  currency: "usd",
  source: "tok_visa", # obtained with Stripe.js
  description: "My First Test Charge (created for API docs)"
}, {
  idempotency_key: "MLiPkjS2LabYa98k"
})
```

drewscott:
Added pull request #191

excid3:
This is now done using the built-in idempotency key support from the Stripe gem.

d7c9291

PR #286, opened 2021-02-26
Stripe idempotency_key support

ryanfb:
This is an update of the PR in #191, which fixes the client-side Stripe idempotency issue described in #188. I've also added passing tests for what's expected with repeated `idempotency_key` requests. Note that the `stripe` gem doesn't raise an exception for repeated requests with the same `idempotency_key`, though the duplicate request is ignored on Stripe's end if it's already been processed. In the case of subscriptions, I added a check that the subscription doesn't already exist so we don't duplicate it on our end. For charges, the existing behavior already works correctly with the `idempotency_key` added.


ryanfb:
FWIW tests for this pass in local development, but I'm unsure the best way to add the VCR cassettes without disclosing my Stripe test keys while still having it pass in the CI environment, and I'm also unsure how to create named API IDs on my corresponding Stripe account like 'small-monthly' (instead of e.g. `price_hA5hVa7uE`) so I have to rename those in the tests before/after running the tests.

excid3:
The Stripe ruby gem automatically handles idempotency keys if you specify retries, so I've added that with a default (which you probably always want). https://github.com/pay-rails/pay/commit/d7c929190a3a7667f3db36da759d7feec5aac3e9

Maybe long-term there should be a way for overriding the idempotency key, but this should be good enough.

PR #288, opened 2021-02-27
Add CSP note to Stripe SCA Payment confirmations

dzacarias:
When apps have a Content Security Policy in place, the domains for the Stripe and Vue scripts, styles and frames should be added to the CSP in order for the payment confirmation pages to work.

Having this note in the README would be helpful in avoiding problems after deployment


excid3:
Thanks @dzacarias 👍

PR #294, opened 2021-03-09
Stripe Connect Support

excid3:
- [x] Adds `stripe_account` to Billable, Charge, and Subscription models
- [x] Adds `Pay::Stripe.transfer` method 
- [x] Direct charges
- [x] Destination charges
- [x] Separate/payout charge & subscription support
- [x] Transfers
- [x] Account lookup
- [x] Connect Webhooks
- [x] Save currency on Pay::Charge
- [x] Save application_fee_amount on Charge (possibly application_fee_percent on Pay::Subscription)
- [x] Associate Pay::Charge with Pay::Subscription using `charge.invoice.subscription` from Stripe::Charge (nice to have)


donnfelker:
Hey @excid3, is there anything else needed to help push this over the finish line? Really looking forward to this.

excid3:
A decent bit left:

1. Onboarding Stripe Connect accounts with AccountLink (been working on this)
2. Adding Payout, Transfer, etc models
3. Test it all

johnchambers:
Testing and I can charge for marketplace payments with USD or EUR which is great. 👍 

I noticed there is no reference to the currency in the PayCharge table however.
Should this be added as an extra data attribute or added as a column in the PayCharge table?

I think adding as a column would be really helpful for mailers to reflect the currency paid and also for the subscription invoices at a later point.

johnchambers:
Recording the application_fee_amount in the PayCharge table would also be important.  

As platforms would need a way to show merchants how much they were charged in fees each month.

excid3:
Agreed on those. 👍 

I think column and application_fee_amount make sense to have as dedicated columns so they're a bit easier to query against.

johnchambers:
So to get Stripe Connect webhooks working alone with the standard Stripe account webhooks do we pass the 2 signing secrets as an array like so?

```
stripe:
    private_key: xxxx
    public_key: yyyy
    signing_secret: ["whsec_stripe_account_signing_secret", "whsec_stripe_connect_signing_secret"]
```

I've tried this out in development and it seems to be working without any other changes.

johnchambers:
Hi @excid3 Just setting up Stipe Express onboarding. I have account_links and login_links working. But how does `onboarding_complete` get set against the merchant account? Is this a value Stripe passes back or do we need to set it ourselves when we are happy the account has enough details added? For example `charges_enabled` and capabilities active.

excid3:
> But how does `onboarding_complete` get set against the merchant account? Is this a value Stripe passes back or do we need to set it ourselves when we are happy the account has enough details added? For example `charges_enabled` and capabilities active.

I don't believe I finished that.

From [the Stripe docs](https://stripe.com/docs/connect/express-accounts#handle-users-not-completed-onboarding):

> A user that’s redirected to your return_url might not have completed the onboarding process. Use the /v1/accounts endpoint to retrieve the user’s account and check for charges_enabled. If the account isn’t fully onboarded, provide UI prompts to allow the user to continue onboarding later. The user can complete their account activation through a new account link (generated by your integration). You can check the state of the details_submitted parameter on their account to see if they’ve completed the onboarding process.

It seems like `charges_enabled` and `details_submitted` are somewhat redundant? I need to confirm with Stripe on that.

johnchambers:
Thanks for the info Chris! I think I'll set the `onboarding_complete` attribute based on charges_enabled or `capabilities: { "card_payments": "active", "transfers": "active" }` for now until I learn more.

test/dummy/db/migrate/20170205000000_create_users.rb
```diff
@@ -11,5 +11,11 @@ def change
       t.string :name
       t.references :owner, polymorphic: true
     end
+
+    create_table :accounts do |t|
+      t.string :email
+      t.string :merchant_processor
```
williamkennedy, line 17: Hey Chris, 

I had a quick play around with this today.

Can we add this to the billable migration? Users would have to add this column to their billable model but there is no indication about where to do that...

Perhaps adding it to the initial pay:install migration would help.

excid3: I wasn't sure if we should require a Merchant to also be Billable, that's why it's separate right now. It seems logical that a Merchant would also be Billable, so maybe it's not necessary to separate them out.

PR #296, opened 2021-03-14
Add paddle environment variable

nm:
Pass Paddle environment variable to PaddlePay.


excid3:
Thanks as always @nm 🙏

PR #301, opened 2021-04-08
Fix error div Id reference

bbonamin:
Just above this function, the div to display errors has a different id, which will be fixed by this update.


excid3:
Thanks!

PR #302, opened 2021-04-09
Add passthrough fallback for paddle payment succeeded webhook

nm:
In my staging tests I realized that although the payment succeeded webhook is fired after the subscription created webhook, it is possible that the payment succeeded webhook is handled first. That means it is not guaranteed that there is already a paddle processor id saved in the database.
For this scenario the billable should be extracted from the passthrough parameter (similar to the subscription created webhook).


excid3:
Good catch @nm 👍

PR #310, opened 2021-04-19
Use bulk: true when running change_table migrations

doutatsu:
As I was running these migration on my personal project, I've noticed that `change_table` is being run without `bulk: true`. Without that flag, all the new columns would be added one by one, which can be significantly slower on large tables. Instead, this will group them into one SQL query.

References:
- https://aaronlasseigne.com/2012/06/05/faster-activerecord-migrations-using-change-table-bulk/
- https://blog.toshima.ru/2020/01/02/rails-migration-bulk-true.html#migration-with-bulk-option
- https://makandracards.com/makandra/3303-bulk-change-multiple-table-rows-in-a-migration

### DIFF
--- a/lib/generators/active_record/templates/migration.rb
+++ b/lib/generators/active_record/templates/migration.rb
@@ -2,7 +2,7 @@
 
 class AddPayBillableTo<%= table_name.camelize %> < ActiveRecord::Migration<%= migration_version %>
   def change
-    change_table :<%= table_name %> do |t|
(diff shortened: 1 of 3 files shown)


excid3:
Looks good to me. 👍 

I don't think there are any downsides to doing the `bulk: true` option?

doutatsu:
@excid3 I think there are some edge-cases when rolling back particular migrations (e.g. when you are setting a `change_column_null` I think), but completely safe in this instance here, where you are just adding columns

Issue #282, opened 2021-02-23
Stripe subscriptions through Checkout sometimes marked "incomplete"


ryanfb:
Testing with new users in my test environment, I've noticed that even with the exact same code and checkout steps, using Stripe Checkout sometimes results in subscriptions which are marked "incomplete", while other times they are correctly(?) marked "active". I think this may be due to race conditions in the webhook callbacks or non-deterministic ordering in the way Stripe calls them, but I'm not sure.

An example of the logs for `stripe listen` on a customer being created and subscribing through Stripe Checkout, with the resulting pay `subscription` being marked as "active":
```
2021-02-23 10:39:49   --> customer.created [evt_1IO2zIBuGOrcp249WQBJ58tj]
2021-02-23 10:39:49  <--  [200] POST http://localhost:3000/pay/webhooks/stripe [evt_1IO2zIBuGOrcp249WQBJ58tj]
2021-02-23 10:40:18   --> checkout.session.completed [evt_1IO2zmBuGOrcp249bXOPacTU]
2021-02-23 10:40:18  <--  [200] POST http://localhost:3000/pay/webhooks/stripe [evt_1IO2zmBuGOrcp249bXOPacTU]
2021-02-23 10:40:18   --> charge.succeeded [evt_1IO2zmBuGOrcp249ef3MVe64]
2021-02-23 10:40:18   --> payment_method.attached [evt_1IO2zmBuGOrcp249Qy6BF3gE]
2021-02-23 10:40:18  <--  [200] POST http://localhost:3000/pay/webhooks/stripe [evt_1IO2zmBuGOrcp249ef3MVe64]
2021-02-23 10:40:18   --> invoice.created [evt_1IO2zmBuGOrcp249MZ2cHow4]
2021-02-23 10:40:18  <--  [200] POST http://localhost:3000/pay/webhooks/stripe [evt_1IO2zmBuGOrcp249MZ2cHow4]
2021-02-23 10:40:18   --> customer.subscription.created [evt_1IO2zmBuGOrcp249QLXlFKQY]
2021-02-23 10:40:18  <--  [200] POST http://localhost:3000/pay/webhooks/stripe [evt_1IO2zmBuGOrcp249QLXlFKQY]
2021-02-23 10:40:18  <--  [200] POST http://localhost:3000/pay/webhooks/stripe [evt_1IO2zmBuGOrcp249Qy6BF3gE]
[...]

PR #329, opened 2021-05-10
Fix stripe subscription webhook order issue

hoppergee:
I try to do a fix for this issue https://github.com/pay-rails/pay/issues/282#issuecomment-836977081


hoppergee:
Since it will be fixed in #339. This PR can be close I think.

PR #333, opened 2021-05-12
stripe-connect webhooks

williamkennedy:
Hey Chris,

So I had an itch and wanted to give this a go at trying to integrate the webhooks myself. 

It seems stripe use the charges_enabled to check if the merchant is fully unboarded so I have opted for just one account update but I would like to add it to more update the pay_data value accordingly...

https://stripe.com/docs/connect/express-accounts#handle-users-not-completed-onboarding

This allow so helps with [Issue 178](https://github.com/pay-rails/pay/issues/178)


lib/pay/merchant.rb
```diff
@@ -2,6 +2,17 @@ module Pay
   module Merchant
     extend ActiveSupport::Concern
 
+    # Keep track of which Merchant models we have
+    class << self
```
williamkennedy, line 6: This was copied from the billiable class. it probably would be good to abstract even though I don't think it will change much

excid3: Yeah, I just ripped this out of Madmin and I'm wondering if we may want to do the same here.

This is nice, but only is reliable if you've eager loaded your whole app. It may make more sense to have a configuration of `merchant_models = ["User", "Account"]`

We can roll with this for now though.

PR #334, opened 2021-05-14
add README note on owner_id column format

mcansky:
both pay_subscriptions and pay_charges need a little tweak as commented in the related migrations (*_create_pay_subscriptions.pay.rb, *_create_pay_subscriptions.pay.rb) :

```
  # Some Billable objects use string as ID, add `type: :string` if needed
  t.references :owner, polymorphic: true
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ To add the migrations to your application, run the following migration:
 
 `bin/rails pay:install:migrations`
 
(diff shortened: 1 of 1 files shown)


mcansky:
:beers: @excid3

excid3:
Thanks @mcansky 🙏

PR #364, opened 2021-06-23
Properly process subscription renewels from stripe

osheroff:
pulling in v2.7.1 to my code.  There's a problem with processing stripe subscriptions that are Renewed after a user cancels them -- because of the way the sync logic worked we weren't properly bringing them "back from the dead".

There's a lot of different fields in the slack api around subscriptions ending/cancelling/etc -- I'm not sure I understand them all, so I hope I'm getting this right.  I'm honestly coding just a little blind. 

It also feels as though we could *probably* just trust the `cancel_at` field in the slack api (without bothering with `cancel_at_period_end` etc) but again I'm not super-familiar with the intimates of the api.

### DIFF
--- a/lib/pay/stripe/subscription.rb
+++ b/lib/pay/stripe/subscription.rb
@@ -37,17 +37,15 @@ module Pay
           trial_ends_at: (object.trial_end ? Time.at(object.trial_end) : nil)
         }
 
-        if object.ended_at
+        attributes[:ends_at] = if object.ended_at
           # Fully cancelled subscription
-          attributes[:ends_at] = Time.at(object.ended_at)
+          Time.at(object.ended_at)
         elsif object.cancel_at
           # subscription cancelling in the future
-          attributes[:ends_at] = Time.at(object.cancel_at)
+          Time.at(object.cancel_at)
         elsif object.cancel_at_period_end
           # Subscriptions cancelling in the future
-          attributes[:ends_at] = Time.at(object.current_period_end)
-        else
-          attributes[:ends_at] = nil
+          Time.at(object.current_period_end)
         end
 
         # Update or create the subscription
--- a/lib/pay/stripe/subscription.rb
+++ b/lib/pay/stripe/subscription.rb
@@ -36,12 +36,15 @@ module Pay
           stripe_account: owner.stripe_account,
           trial_ends_at: (object.trial_end ? Time.at(object.trial_end) : nil)
         }
-
-        # Subscriptions cancelling in the future
-        attributes[:ends_at] = Time.at(object.current_period_end) if object.cancel_at_period_end
-
-        # Fully cancelled subscription
-        attributes[:ends_at] = Time.at(object.ended_at) if object.ended_at
+        if object.ended_at
+          # Fully cancelled subscription
+          attributes[:ends_at] = Time.at(object.ended_at)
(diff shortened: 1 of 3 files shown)


excid3:
Can you add a test case that shows the issue?

lib/pay/stripe/subscription.rb
```diff
@@ -36,12 +36,15 @@ def self.sync(subscription_id, object: nil, name: Pay.default_product_name)
...
-        attributes[:ends_at] = Time.at(object.ended_at) if object.ended_at
+        if object.ended_at
+          # Fully cancelled subscription
+          attributes[:ends_at] = Time.at(object.ended_at)
+        elsif object.cancel_at_period_end
+          # Subscriptions cancelling in the future
+          attributes[:ends_at] = Time.at(object.current_period_end)
+        elsif object.cancel_at.nil?
+          attributes[:ends_at] = nil
+        end
```
excid3, line 47: I'm thinking this might need to be changed so we handle the `cancel_at` column properly.

```ruby
elsif object.cancel_at
  attributes[:ends_at] = Time.at(object.cancel_at)
else
  attributes[:ends_at] = nil
end
```

This would then mark the subscription `ends_at` anytime there's an ended_at, cancel_at, or cancel_at_period_end. Resetting to nil if not.

excid3:
This was a good catch @osheroff! Appreciate the help!

I'll cut a release with this fix as soon as we're done.

osheroff:
> Can you add a test case that shows the issue?

yeah.  am I correct that the `vcr` based harnesses are sort of deprecated in favor of just stubbing?

lib/pay/stripe/subscription.rb
osheroff: that was what I was thinking too, but didn't know if there were special meanings here or there.

osheroff:
> Another is a subscription with cancel_at set to a value and a webhook that updates that to nil. I actually never used this field, so it's good we add support for it.

yeah.  I'm not sure how `cancel_at` gets set in stripe other than subscription cancellation (at the end of the billing cycle), but it seems relatively authorative.

excid3:
The VCR stuff is hard to deal with recording and updating since it needs a real Stripe test account to run against. IDs and everything change. That's why I am leaning towards stubbing instead so it's more reliable.

I think they all serve different purposes:

`canceled_at` - subscription is canceled and cannot be resumed
`cancel_at` - marks as subscription as cancelling anytime in the future
`cancel_at_period_end` - shortcut for setting the cancel_at to the end of the billing period. I would imagine this sets `cancel_at`

If it does set `cancel_at`, then we could probably remove the logic for `cancel_at_period_end` since it wouldn't be necessary. Would need to test that to make sure though.

osheroff:
ok @excid3 I pushed some tests and worked the logic a bit more. 

I checked the stripe documentation but don't really have much visibility into whether `cancel_at` is *always* set, or whether there's corner cases where `cancel_at_period_end` might be set without `cancel_at` being set... so I left both checks in, because I'm a coward.  Especially when it comes to billing :)

osheroff:
wasn't a giant fan of the coding style standardrb dictated but it insisted, so I relented.

osheroff:
ping

excid3:
Thanks for this @osheroff 👍

Issue #371, opened 2021-07-03
scary race condition in stripe webhooks


osheroff:
I've only encountered this in development mode, but it appears that stripe has made their webhooks usability even worse these days by now firing off three events (in my case, `payment_intent.succeeded`, `charge.succeeded`, and `invoice.paid`) in parallel.  Since we aren't enforcing any uniqueness constraints, these three webhooks calls raced against each other, and I ended up with 3 identical subscription rows in the DB.  

Any ideas on how to fix?  a unique index on (`processor`, `processor_id`) in subscriptions might fix this case but I'm not sure how it would ripple out to the system.

Other than that we could do some transaction select-for-update stuff.  maybe.

excid3:
That's a great question. We do need a uniqueness validation probably on both the Pay::Charge and Pay::Subscription models.

I recently changed the webhooks to simply trigger a sync. Sometimes the webhooks come in out of order, so occasionally it would update a record like a Subscription with old state. Now the webhooks just trigger an API request to get the latest information every time.

The uniqueness validation should take care of creating duplicates, but it might be a good idea to lock the record as well for updates.

Think that would handle most situations?

osheroff:
not clear, are you talking about doing a rails uniqueness validation or a db-level constraint?  
seems likely that doing it as just a rails validation would just be subject to the same race condition.

I've never been clear in my mind how locking a yet-to-be-inserted record works.  does AR have some kind of pattern?

excid3:
Both, for sure. The database level constraint will be the only reliable way to handle that, and the Rails validation can just help a bit.

I don't think you can lock a yet-to-be-inserted record. But locking the row would be useful for updating subscription state.

osheroff:
yeah @excid3 that all sounds good.  You want to do the code changes or me?

excid3:
I leave for vacation tomorrow, so if you've got time, that would be awesome. 👍

osheroff:
hah i leave for vacation tomorrow too.

excid3:
LOL, hilarious timing. I may take a quick stab at it today if I get some free time.

excid3:
@osheroff I'm back from vacation and was poking at this again. I merged without the retry but was playing with that too.

Does this make sense for retries?

```ruby
      def self.sync(charge_id, object: nil, try: 0, retries: 5)
        object ||= ::Stripe::Charge.retrieve(id: charge_id)
        owner = Pay.find_billable(processor: :stripe, processor_id: object.customer)
[...]

PR #377, opened 2021-07-14
Validate uniqueness of Charges and Subscriptions

excid3:
Now that we're listening to more webhooks that are semi-redundant, we need to make sure we're not saving duplicates. This adds a db constraint and validations.

We want to make sure that:

* Webhooks creating the same record handle this safely.
* Webhooks updating the same record lock for changes.

#371


excid3:
@osheroff I think this covers everything. I added locks when updating, although I'm not entirely sure how useful they will be.

osheroff:
looks good!

still a touch racy in a weird way (if 3 webhooks race to create the row, the api fetch from the first one wins and the other two fail, leaving a maybe-out-of-date object), but it's definitely better than creating multiple subscription rows.  I'd say we try it in production, and if stripe sends fast enough to race creation a bunch of the time maybe we could rescue-retry-update the failed ones?

excid3:
Oh yeah, that's what I was forgetting. I had thought about doing the rescue + short sleep + retry for those. I think that would make sense and would trigger the lock for the retries which should handle it as best we can.

Otherwise we'd have to queue things and process them one-by-one, which sounds awful.

Issue #174, opened 2020-06-24
Inconsistent assignment of subscription `processor_plan`


archonic:
I noticed some odd behaviour when writing request tests that call `swap` on the user subscription.

After setting up my subscribed user which is testing against live Stripe,`user.subscription(name: "example").processor_plan` is a JSON blob:

```
{
  "id": "price_1GvnFzFpMcvkWhcFuMrM9m9c",
  "object": "plan",
  "active": true,
  "aggregate_usage": null,
  "amount": 900,
  "amount_decimal": "900",
  "billing_scheme": "per_unit",
  "created": 1592584799,
  "currency": "usd",
  "interval": "month",
  "interval_count": 1,
  "livemode": false,
  "metadata": {
  },
  "nickname": null,
  "product": "prod_HU72MDZXgDBkYM",
  "tiers": null,
  "tiers_mode": null,
  "transform_usage": null,
  "trial_period_days": 14,
  "usage_type": "licensed"
}
```

After calling `user.subscription(name: "example").swap("pm_xxx")`, the `processor_plan` attribute is just the ID string (`price_1GvnFFFpMcvkWhcFleWYnqs5`).

As far as I can tell, it looks like the intention was for `processor_plan` to just be the ID string. If that's true, it looks like this is the culprit:
https://github.com/pay-rails/pay/blob/master/lib/pay/billable.rb#L135

Happy to submit a PR, just wanted to ask what the intention was for that attribute first.

archonic:
This was happening because I was passing in the whole Stripe::Plan object instead of just the ID. My mistake, but perhaps there should be an argument error if `plan` is not a `String`?

excid3:
I think that'd be a great PR. ;P

PR #379, opened 2021-07-19
raises ArgumentError when the argument passed to swap method is not a String…

A7madXatab:
According to https://github.com/pay-rails/pay/issues/174#issue-644277572. sometimes users mistake the argument for `swap` to be an instance of an Object. But, `swap` only accepts a string that corresponds to the Plan/Price etc.

And according to https://github.com/pay-rails/pay/issues/174#issuecomment-650606862 this is a desired behaviour.

This PR makes all `swap` methods(including fake_processor) to raise an exception when the argument passed to them is not an instance of `String`

### DIFF
--- a/test/pay/stripe/subscription_test.rb
+++ b/test/pay/stripe/subscription_test.rb
@@ -48,11 +48,13 @@ class Pay::Stripe::SubscriptionTest < ActiveSupport::TestCase
   end
 
   test "it will throw an error if the passed argument is not a string" do
-    @user.processor_id = nil
+    @user.processor = stripe
     @user.card_token = "pm_card_visa"
+    @user.subscribe
 
-    plan_hash = {}
-    assert_raise(Pay::Stripe::Error) { @user.subscribe(name: "default", plan: plan_hash) }
+    assert_raises Pay::Stripe::Error do
+      @user.swap({invalid: :object})
+    end
   end
 
   private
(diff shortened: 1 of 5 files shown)


excid3:
Thanks @A7madXatab 👍

excid3:
Looks like this just needs a `bundle exec standardrb --fix` on it.

A7madXatab:
> bundle exec standardrb --fix

will be doing so as soon as possible.

A7madXatab:
> bundle exec standardrb --fix

@excid3 all good, good to go!.

test/pay/stripe/subscription_test.rb
```diff
@@ -47,6 +47,14 @@ class Pay::Stripe::SubscriptionTest < ActiveSupport::TestCase
     assert_not_nil pay_subscription.ends_at
   end
 
+  test "it will throw an error if the passed argument is not a string" do
+    @user.processor_id = nil
+    @user.card_token = "pm_card_visa"
+
+    plan_hash = {}
+    assert_raise(Pay::Stripe::Error) { @user.subscribe(name: "default", plan: plan_hash) }
```
excid3, line 55: This should test the `swap` method and it currently doesn't.

Should be more like
```
@user.processor = stripe
@user.card_token = "pm_card_visa"
@user.subscribe
assert_raises ArgumentError do
  @user.swap({invalid: :object})
end
```

excid3:
Rewrote the test and merged this. Thanks @A7madXatab

A7madXatab:
> Rewrote the test and merged this. Thanks @A7madXatab

sorry for the inconveniences made in my commits @excid3 , thanks for merging this!.

excid3:
You're fine! The tests for this gem are a pain so no worries. 😅

PR #393, opened 2021-08-04
Pay v3

excid3:
* Migrates `processor` and `processor_id` from models to `Pay::Customer` model
* Replaces `include Pay::Billable` with `pay_customer` method
* Replaces `include Pay::Merchant` with `pay_merchant` method
* Changes `Pay::Charge` to associate with `Pay::Customer` instead of `owner{polymorphic}`


excid3:
Use `update!` when updating customers and things because we don't want to lose the processor ID.

PR #399, opened 2021-08-15
Improving the readme

guillaumebriday:


excid3:
@guillaumebriday thanks!

I should give you a heads up, we've got a big PR for Pay v3 (see the branch) that is almost ready to ship and I'd recommend using that right now so you won't have to do the migration in your app. Basically moving models around so it's much more smooth to use.

guillaumebriday:
Thanks for the merge!

I need to take a look at v3, maybe we should write an upgrade guide if needed. I'm currently running v2 in production so I will try to upgrade to v3 when it will be merged!

excid3:
Ah okay! So you're in my boat too. Upgrading is going to be interesting. I will have to get your feedback on the upgrade migration. It is a little tricky. I'll be uploading that soon. Trying to test it on Jumpstart Pro at the moment.

PR #401, opened 2021-08-17
Fix typo in method

christopheragnus:
Typo in validation.


christopheragnus:
Mayeb the method should be renamed? It's pretty confusing to name it the same as the ActiveRecord attribute.

excid3:
@christopheragnus oops, you are correct! Thanks!

excid3:
This is cleaned up in v3 which I'm just getting ready to release.

PR #403, opened 2021-08-20
Cancel active subscriptions when Owner is deleted

stevepolitodesign:
@excid3 before I get too far, I want to make sure I'm going in the right direction. I wrote a failing test as a guide.

**Questions**
- Should I even be concerned about when the `Owner` is deleted, or is it enough to just test when the `Pay::Customer` is deleted?
- Should the test be in the Subscription namespace for each Payment Processor namespace? Or should it be in `Pay::Billable::SyncCustomer::Test`?

### DIFF
--- a/app/models/pay/subscription.rb
+++ b/app/models/pay/subscription.rb
@@ -15,6 +15,9 @@ module Pay
     scope :incomplete, -> { where(status: :incomplete) }
     scope :past_due, -> { where(status: :past_due) }
 
+    # Callbacks
+    before_destroy :cancel_now!, if: :active?
+
     # TODO: Include these with a module
     store_accessor :data, :paddle_update_url
     store_accessor :data, :paddle_cancel_url
--- a/lib/pay/attributes.rb
+++ b/lib/pay/attributes.rb
@@ -12,6 +12,8 @@ module Pay
         has_many :charges, through: :pay_customers, class_name: "Pay::Charge", dependent: :destroy
         has_many :subscriptions, through: :pay_customers, class_name: "Pay::Subscription", dependent: :destroy
         has_one :payment_processor, -> { where(default: true) }, class_name: "Pay::Customer", as: :owner, inverse_of: :owner
+
+        after_commit :cancel_active_subscriptions!, on: :destroy
       end
 
       # Changes a user's payment processor
@@ -54,5 +56,12 @@ module Pay
         end
       end
     end
+
+    private
+
+    def cancel_active_subscriptions!
+      byebug
+    end
+
   end
 end
--- a/lib/pay/attributes.rb
+++ b/lib/pay/attributes.rb
@@ -12,8 +12,6 @@ module Pay
         has_many :charges, through: :pay_customers, class_name: "Pay::Charge"
         has_many :subscriptions, through: :pay_customers, class_name: "Pay::Subscription"
         has_one :payment_processor, -> { where(default: true) }, class_name: "Pay::Customer", as: :owner, inverse_of: :owner
-
-        before_destroy :cancel_active_subscriptions!
       end
 
       # Changes a user's payment processor
@@ -55,12 +53,6 @@ module Pay
           end
         end
       end
-    end
-
-    private
-
-    def cancel_active_subscriptions!
-      byebug
     end
 
   end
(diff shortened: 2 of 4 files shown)


excid3:
1. You would want to test against the owner destroy because that's the situation when a User gets deleted in your app.
 
That does bring up the question of what happens to the Pay::Customer records though... if we deleted the User, that would leave `owner:polymorphic` blank on the Pay::Customer which isn't valid. Which means you'd have to delete them, and then the Pay::Subscription and Pay::Charges too. HMMM! 🤔 

I guess this would be fine, but we just would need to make sure to `cancel_now!` the subscription(s) before destroy.

2. For now, let's put it in this file since it's closest associated with the Pay::Customer / Pay::Subscription models. https://github.com/pay-rails/pay/blob/master/test/models/pay/subscription_test.rb

stevepolitodesign:
@excid3 here's an overview of the latest update.

- Added `before_destroy :cancel_now!, if: :active?` to `app/models/pay/subscription.rb`
- Added `dependent: :destroy` to `has_many :pay_customers, class_name: "Pay::Customer", as: :owner, inverse_of: :owner` in `lib/pay/attributes.rb `

excid3:
@stevepolitodesign I think this should cover it pretty well. 👍  

I fixed a couple things standardrb wasn't happy with. A couple tests are failing, but they seem unrelated. Seems like it isn't catching the VCR stubs.

excid3:
Working on the tests that are failing. Almost got those fixed.

I noticed Stripe's `delete` customer webhook does something similar:
>Permanently deletes a customer. It cannot be undone. Also immediately cancels any active subscriptions on the customer.

>Unlike other objects, deleted customers can still be retrieved through the API, in order to be able to track the history of customers while still removing their credit card details and preventing any further operations to be performed (such as adding a new subscription).

Which now brings up the question...should we just soft delete the Pay::Customer? 😅

stevepolitodesign:
I guess if that’s what Stripe does, you might as well make that part of the Pay API.

I know this PR is merged, but is there anything you need from me for that?

Does the ‘dependent:  :destroy’ call need to be removed?

excid3:
Great question...

There's basically 3 situations we have to address

1. Stripe::Customer is deleted (in Stripe). We listen to the webhook and soft delete the Pay::Customer. When a Stripe::Customer is deleted, it cancels their subscriptions and removes all their payment methods. I updated Pay to handle this.
2. A developer deletes a Pay::Customer record. We should cancel any active subscriptions and then dependent destroy the Pay::Customer's payment methods, charges, and subscriptions.
3. A Pay::Customer's owner is deleted. We should run #2 for each of their Pay::Customers

So basically if a Customer is remotely deleted on the payment processor, we'll soft delete so the user in the app doesn't lose their Charge / Subscription history.

But if a Pay::Customer is explicitly deleted, it's okay to delete the charges and subscriptions (since that's what they're asking for). We won't delete the Customer via the API (because Stripe makes it very hard to find them without the ID).

I think that seems fine. Can you think of any other cases @stevepolitodesign?

excid3:
Austin Miller and I were chatting... maybe we just orphan the Pay::Customer records (remove the `dependent: :destroy`) and that way we don't lose that when deleting a Account/User record. 

Then the only time we'd destroy charges and subscriptions is if someone explicitly deleted the Pay::Customer.

That's probably the safest option. It would leave orphaned Pay::Customers which is probably okay so you don't lose payment history in-app.

stevepolitodesign:
@excid3 if we remove the `dependent: :destroy` call on the `Owner` we should still fire a callback to cancel any `active` `Pay::Subscription` records though, right?

```ruby
# lib/pay/attributes.rb
module Pay
 ...
    included do
      # Remove dependent: :destroy
      has_many :pay_customers, class_name: "Pay::Customer", as: :owner, inverse_of: :owner
      # Once the record is deleted, cancel any active subscriptions
      after_commit :cancel_active_subscriptions!, on: [:update, :destroy]
    end

    private
    
    def cancel_active_subscriptions!
        self.subscriptions.active(&:cancel_now!)
    end     
end     
```

excid3:
Exactly. I think that's probably safer so we don't lose the Pay::Charge records.

Just need to confirm if deleting the owner will break the Pay::Customer validations I guess.

stevepolitodesign:
@excid3 cool, I can make another PR for this unless you're already working on this refactor.

excid3:
I just did it. Although it's worth reviewing: c149cd9

Now we have the two situations:

* If an active Pay::Subscription record is destroyed, we'll cancel it.
* If a Pay::Customer's owner is destroyed, we'll cancel any active subscriptions.

That all make sense?

PR #405, opened 2021-08-24
Add scopes to Pay::Charge and Pay::Subscription to account for Pay::Customer deleted_at column 

stevepolitodesign:
Now that we're soft deleting `Pay::Customer` records I figured it might be helpful to add some scopes on the `Pay::Charge` and `Pay::Subscription` models. This way one could query for records that are tied to a deleted customer. 

```ruby
Pay::Charge.with_active_customer
# => #<ActiveRecord::Relation []>
Pay::Charge.with_deleted_customer
# => #<ActiveRecord::Relation []>
Pay::Subscription.with_active_customer
# => #<ActiveRecord::Relation []>
Pay::Subscription.with_deleted_customer
# => #<ActiveRecord::Relation []>
```

Also, add helper methods to `Pay::Customer`

```ruby
Pay::Customer.first.active?
# => true
Pay::Customer.first.deleted?
# => false
```

### DIFF
--- a/test/models/pay/charge_test.rb
+++ b/test/models/pay/charge_test.rb
@@ -47,12 +47,11 @@ class Pay::Charge::Test < ActiveSupport::TestCase
   test "with_deleted_customer scope" do
     charge = pay_charges(:stripe)
     customer = charge.customer
-    charges = Pay::Charge.with_deleted_customer
 
-    refute_includes charges, charge
+    refute_includes Pay::Charge.with_deleted_customer, charge
     customer.update(deleted_at: Time.now)
 
-    assert_includes charges, charge
+    assert_includes Pay::Charge.with_deleted_customer, charge
   end
 
   private
--- a/test/models/pay/charge_test.rb
+++ b/test/models/pay/charge_test.rb
@@ -34,4 +34,30 @@ class Pay::Charge::Test < ActiveSupport::TestCase
     charge.update(metadata: metadata)
     assert_equal metadata, charge.metadata
   end
+
+  test "with_active_customer scope" do
+    charge = pay_charges(:stripe)
+    customer = charge.customer
+    charges = Pay::Charge.with_active_customer
+
+    assert_includes charges, charge
+    customer.update(deleted_at: Time.now)
+
+    refute_includes charges, charge
+  end
+  
+  test "with_deleted_customer scope" do
+    charge = pay_charges(:stripe)
+    customer = charge.customer
+    charges = Pay::Charge.with_deleted_customer
+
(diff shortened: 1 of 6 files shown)


test/models/pay/subscription_test.rb
```diff
@@ -118,6 +118,28 @@ class Pay::Subscription::Test < ActiveSupport::TestCase
...
+
+  test "with_deleted_customer scope" do
+    subscription = create_subscription
+    subscriptions = Pay::Subscription.with_deleted_customer
+
+    refute_includes subscriptions, subscription
+
+    @pay_customer.update!(deleted_at: Time.now)
+
+    assert_includes subscriptions, subscription
```
excid3, line 140: You're saving the results of the query, so the second assert here doesn't re-run the query.

Instead you want two fresh queries:
```ruby
    refute_includes Pay::Subscription.with_deleted_customer, subscription

    @pay_customer.update!(deleted_at: Time.now)

    assert_includes Pay::Subscription.with_deleted_customer, subscription
```

test/models/pay/charge_test.rb
```diff
@@ -34,4 +34,28 @@ class Pay::Charge::Test < ActiveSupport::TestCase
...
+
+  test "with_deleted_customer scope" do
+    charge = pay_charges(:stripe)
+    customer = charge.customer
+    charges = Pay::Charge.with_deleted_customer
+
+    refute_includes charges, charge
+    customer.update(deleted_at: Time.now)
+
+    assert_includes charges, charge
```
excid3, line 57: Same as the Subscription comment, you want to query this twice (not saving the results to a variable), otherwise it won't see any changes.

```ruby
    refute_includes Pay::Charge.with_deleted_customer, charge
    customer.update(deleted_at: Time.now)
    assert_includes Pay::Charge.with_deleted_customer, charge
```

It also makes the test more clear I think.

test/models/pay/subscription_test.rb
```diff
@@ -118,6 +118,28 @@ class Pay::Subscription::Test < ActiveSupport::TestCase
...
+  test "with_active_customer scope" do
+    subscription = create_subscription
+    subscriptions = Pay::Subscription.with_active_customer
+
+    assert_includes subscriptions, subscription
+
+    @pay_customer.update!(deleted_at: Time.now)
+
+    refute_includes subscriptions, subscription
+  end
```
excid3, line 130: Same with this test. 👍

excid3 (changes requested):
A couple adjustments should fix those tests for you. 👍

excid3 (approved):

PR #416, opened 2021-09-03
Allow UserMailer delivery to be configured.

### LINKED ISSUE #234: Configure which emails we want to send or not
Thanks for the awesome library!

Would be great if we could configure which emails we want to send or not, eg in my case I don't want to send a renewal email but still want all the others.

stevepolitodesign:
I'm still exploring how to do this, but I'm not entirely sure how to namespace it such that `Pay.emails.some_method` can be called correctly.

The following works, but I doubt it's the Ruby way because the tests are failing intermittently.  

```ruby
module Pay
 ...
  module Emails
    mattr_accessor :receipt
    @@receipt = true

    mattr_accessor :refund
    @@refund = true

    mattr_accessor :subscription_renewing
    @@subscription_renewing = true

    mattr_accessor :payment_action_required
    @@payment_action_required = true
  end

  def self.emails
    Emails
  end
end
```

Once I get that sorted out, we'll need to determine if we even want to allow someone to configure deliverability for the `receipt`, `refund` and `payment_action_required` mailer actions.

---

Closes #234

### DIFF
--- a/lib/pay.rb
+++ b/lib/pay.rb
@@ -48,6 +48,24 @@ module Pay
 
   mattr_accessor :routes_path
   @@routes_path = "/pay"
+  
+  module Emails
+    mattr_accessor :receipt
+    @@receipt = true
+
+    mattr_accessor :refund
+    @@refund = true
+
+    mattr_accessor :subscription_renewing
+    @@subscription_renewing = true
+
+    mattr_accessor :payment_action_required
+    @@payment_action_required = true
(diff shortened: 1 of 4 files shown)


excid3:
An alternative would be to use an OpenStruct for the configuration, similar to how we handle JSON objects in the API wrapper gems.

Not sure if it adds many pros/cons, but it'd be flexible at least?

PR #421, opened 2021-09-05
Fixes link for adding a payment processor

BilalBudhani:


excid3:
Thanks @BilalBudhani 🔥

PR #426, opened 2021-09-08
Correct syntax for setting `payment_processor`

rodreegez:
I _think_ the currently documented syntax is from Pay2, while Pay3 has subtly updated the syntax to no longer be an assignment to `processor`, but instead an argument passed to `set_payment_processor`. Is that right?


excid3:
Good catch! Thanks @rodreegez! I appreciate the help finding those missing pieces in the docs. 💪

PR #428, opened 2021-09-09
Use `pay_customer_name` if available

rodreegez:
In order to allow setting a specific name to be used as the
`pay_customer_name`, check for the existance of `pay_customer_name` on
the `owner` and use that if present. Else, try `name` and then
`first_name` + `last_name` as before.


excid3:
This is fantastic. I was thinking about this a week or two ago and it totally slipped my mind. Glad you caught it. 👍

PR #430, opened 2021-09-09
Update 8_customizing_models.md

the-anthony-s:
`module Charge` will cause the following error: `expected file /Users/tony/Documents/fisnacks/app/models/concerns/charge_extensions.rb to define constant ChargeExtensions`


excid3:
Thank you sir! 🙏

PR #437, opened 2021-09-14
Fix Fake Processor link in README

onerinas:


excid3:
Thank you @onerinas 👍

PR #452, opened 2021-10-04
fix 6.1 rails reference in 2.x

Zhong-z:
```
rake aborted!
ArgumentError: Unknown migration version "6.1"; expected one of "4.2", "5.0", "5.1", "5.2"
```

Fix Rails 6.1 reference in 2.x when using in <6 Rails app


excid3:
Thanks @Zhong-z 👍

Issue #456, opened 2021-10-15
Configure Processors List


platonv:
The processor setup seems to be conditioned by their respective modules being defined. This can cause issues if someone wants to use one processor through this library, but other processor with the processor's SDK.

I ran into this problem because I was using Stripe SDK for some API calls and then I added pay-rails to handle Paddle payments. Just because I had the stripe module defined, pay-rails was overriding `Stripe.api_key` with nil based on `STRIPE_PRIVATE_KEY` env var.

Stripe uses `STRIPE_API_KEY` for this config so an obvious solution is to just rename the env var, but this could catch people off guard and break production code like in my scenario.

I think a better solution would be to allow for a `payment_processors` config variable that would use a list of symbols to explicitly state which processors should be used. This list should be used in this code:

```
    config.to_prepare do
      Pay::Stripe.setup if defined? ::Stripe
      Pay::Braintree.setup if defined? ::Braintree
      Pay::Paddle.setup if defined? ::PaddlePay

      Pay::Charge.include Pay::Receipts if defined? ::Receipts::Receipt
    end
```

I didn't inspect this repo so well, so if you have any other solution, that would be great too. I could open a PR for this myself too.

PR #464, opened 2021-10-27
Enable processors based on config

platonv:
This attempts to solve  #456


excid3:
We ended up taking a slightly different approach in #574 for this.

PR #465, opened 2021-10-27
Updates webhook routes declaration

zedtux:
This commit only declare payment processor webhook route for installed ones, preventing from having unused webhook routes.


zedtux:
@excid3 would you like to review this PR please?

PR #472, opened 2021-11-04
Remove duplicated "cancel" URL reference

jpemberthy:
Looks like a typo, thanks for the excellent work with `pay` ✨


excid3:
Thanks @jpemberthy 💪

PR #474, opened 2021-11-04
Fix fake processor trial days issue

### LINKED ISSUE #473: `trial_period_days` doesn't work with the fake processor
When using the fake_processor, the following throws an error:
```
    user.payment_processor.subscribe(
      plan: plan,
      name: slug,
      quantity: 1,
      trial_period_days: 14
    )
```

Workaround is to use `trial_ends_at`, but the wiki tells you to use `trial_period_days`

rathboma:
fix #473


rathboma:
Worth noting - when doing this it creates a bunch of files in test/vcr_cassettes that are untracked but not excluded by .gitignore, not sure if that's a bug?

excid3:
Thank you @rathboma 👍

PR #475, opened 2021-11-12
Update UPGRADE.md

clemoun:
I added a reference to your "Pay 3.0 Upgrade for Jumpstart Pro" that gives some interesting infos regarding the two migrations and how to handle them


excid3:
Thanks 👍

PR #478, opened 2021-11-17
Remove binding.irb statement in dummy app

jackspiva:
Something I noticed when looking through the dummy app. I haven't actually used `binding.irb` before, but my impression is that it's a debugging statement similar to pry's `binding.pry` so I assume it just got left in by accident.


excid3:
Yep thanks!

PR #482, opened 2021-11-22
Quantity delete fix

rathboma:
Quantity is not valid as a top level property. We already include it for line items, but stripe checkout breaks if you don't delete it from the top level options.


rathboma:
@excid3 this is a simple fix that simply un-breaks checkout when a quantity is specified.

rathboma:
current workaround is to specify line items as a hash.

excid3:
I figure it'd be better to delete the quantity when we use it instead so I moved the delete there. Thanks!

PR #489, opened 2021-11-30
Add configuration option `generate_receipts` and Documentation Improvements

westonganger:
https://github.com/pay-rails/pay/discussions/488

### DIFF
--- a/docs/2_configuration.md
+++ b/docs/2_configuration.md
@@ -90,19 +90,20 @@ Need to make some changes to how Pay is used? You can create an initializer `con
 
 ```ruby
 Pay.setup do |config|
-  # For use in the receipt/refund/renewal mailers
-  config.business_name = "Business Name"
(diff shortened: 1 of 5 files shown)


docs/2_configuration.md
```diff
@@ -90,19 +90,20 @@ Need to make some changes to how Pay is used? You can create an initializer `con
...
 ```ruby
 Pay.setup do |config|
-  # For use in the receipt/refund/renewal mailers
-  config.business_name = "Business Name"
-  config.business_address = "1600 Pennsylvania Avenue NW"
-  config.application_name = "My App"
-  config.support_email = "helpme@example.com"
-
   config.send_emails = true
+  config.generate_receipts = true
```
excid3, line 94: Is there a use case where you'd have the `receipts` gem installed and _not_ want to include receipts in your emails?

PR #490, opened 2021-11-30
Update 7_webhooks.md

ro31337:
Should be the right link without `docs/` prefix, because when I click it from https://github.com/pay-rails/pay/blob/master/docs/7_webhooks.md - I'm getting 404.


excid3:
Thanks @ro31337! Those darn relative links. 😜

PR #505, opened 2021-12-22
Remove extra currencies

olimart:
No need for gold, silver, and other currencies meant for testing purpose


olimart:
@excid3 if you could please approve running workflow 🙏

excid3:
We're using the Money gem's currency json file, so I'm not too worried about including gold, silver, etc. Just want to make sure this is easy to update.

We can remove it, but it'll likely get added back in the future when we sync changes from the Money gem.

olimart:
@excid3 up to you, I saw that you rely on this file for Jumpstart therefore exposing these options on the UI.
No hard feelings if you decide to close it ;)

excid3:
Yeah, it helps let us render currencies without fully pulling in Money. I think I'll just close this for now.

PR #508, opened 2021-12-22
Add Stripe creds docs, fix a couple doc samples

cjavilla-stripe:


excid3:
Thanks @cjavilla-stripe! 🙏

Issue #376, opened 2021-07-13
Hardcoded payment_method_types


doutatsu:
Correct me if I am wrong, but looking into the source code, `payment_method_types` is hardcoded to `card`. Is there no way to provide additional payment types? If not, is there a chance that ability would be added any time soon?

excid3:
Just pass in that option to override it. 

But yes, currently only cards will be saved. I'll be adding a PaymentMethod model when I get back from vacation so we can store more options.

doutatsu:
Ah, I am blind, you can override them, my bad. Thanks for letting me know!

excid3:
Made some good progress on this in the `v3` branch. Tested all the different options I could from Stripe and Braintree.

cjavilla-stripe:
We can actually _remove_ `payment_method_types` now as it's optional. If it's not passed, the Checkout Session will dynamically present the payment method types that are configured in the Stripe dashboard.

doutatsu:
Yup, I was thinking of Pay when I read the latest update. This really should have been there from day one - will make life so much easier

excid3:
Going to close this now that we've updated the Stripe Checkout helpers to not include them any longer. 👍

clemoun:
Hello,

Sorry for the beginner's question but Stripe is telling me "We've detected that you are not using Automatic Payment Methods" which means (if I'm correct) that `payment_method_types` still defaults to `["card"]` in my app.

I first installed Pay in december on my app and this issue was fixed by @cjavdev in january : how do I get the latest version of Pay in my rails app ?
To this day, my Gemfile says `gem 'pay', '~> 3.0'`

excid3:
Use it from master.

PR #519, opened 2022-01-08
Enable automatic payment methods for Stripe Checkout

cjavdev:
Removes hard coded `payment_method_types` to enable automatic payment methods to be surfaced in Checkout based on the Stripe Dashboard configuration. 

`payment_method_types` can still be passed in, but this will increase conversion by default by accepting more payment methods.

Helps address #376


excid3:
Thanks @cjavdev 🙏

PR #527, opened 2022-01-23
Allow syncing customers by overriding `pay_should_sync_customer?`

nhocki:
Before this change it was possible to trigger the `enqeue_sync_email_job` method by changing the
`pay_should_sync_customer?` but the method would be a no-op unless the email changed.

This pull request makes it possible for someone to override _when_ the customer would be synced by overriding the `pay_should_sync_customer?` method.

The idea behind this is that the customer's `name` will also be synced to Stripe when they change it in our app. We would achieve this by adding this to `models/user.rb`:

```ruby
def pay_should_sync_customer?
  saved_change_to_email? || saved_change_to_name?
end
```

### DIFF
--- a/lib/pay/billable/sync_customer.rb
+++ b/lib/pay/billable/sync_customer.rb
@@ -8,7 +8,7 @@ module Pay
       extend ActiveSupport::Concern
 
       included do
-        after_update :enqeue_sync_email_job, if: :pay_should_sync_customer?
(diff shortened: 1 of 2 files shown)


excid3:
Looks good to me, thanks! 🙏

excid3:
I will once #546 is finished and merged.

PR #539, opened 2022-02-10
Update Paddle docs to add public key encoding

deanpcmad:
I'm currently in the process of setting up an application with Pay & Paddle and I found that there was missing documentation for the Public Key so I've added it in this PR

### DIFF
--- a/docs/paddle/1_overview.md
+++ b/docs/paddle/1_overview.md
@@ -19,3 +19,21 @@ The [Paddle Sandbox](https://developer.paddle.com/getting-started/sandbox) can b
   Paddle.Setup({ vendor: <%= Pay::Paddle.vendor_id %> });
 </script>
 ```
+
+## Paddle Public Key
+
+Paddle uses public/private keys for webhook verification. You can find
+your public key [here for Production](https://vendors.paddle.com/public-key)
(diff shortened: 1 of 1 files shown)


excid3:
I wonder if we should just update Pay to support both base64 encoded and plain text public keys? 

It seems a bit unnecessary to force users to base64 encode the public key when credentials can store multiline strings just fine. I think the reason they suggest base64 encoding is for environment variables or similar things where it's not easy to include multiline strings.

deanpcmad:
Yeah I did wonder why it was only set to accept base64.
I could work on adding that. I guess it would be an idea to support both and then deprecate the base64 version in a newer version?

excid3:
I imagine we'd always want to support both, especially since it's in their documentation.

Maybe we modify the `Pay::Paddle` class to have a `public_key` method that either returns the raw public_key, or decodes the `public_key_base64`

https://github.com/pay-rails/pay/blob/587800a72c8ffaaae72034beb9788c36b801af57/lib/pay/paddle.rb#L38

Then the verifier can skip the base64 decode entirely. https://github.com/pay-rails/pay/blob/587800a72c8ffaaae72034beb9788c36b801af57/lib/pay/paddle/webhooks/signature_verifier.rb#L16

PR #543, opened 2022-02-14
Include metadata when creating customer in Stripe

### LINKED ISSUE #431: Pay Customer metadata
Similar to #428 by @rodreegez, it would be great to support a `pay_customer_metadata` method on Stripe Billable that could sync metadata when a Stripe customer is created / updated.

This would be useful for syncing other information like the ID of the user in your database to Stripe.

excid3:
Adds the ability to define a symbol or Proc that returns metadata when creating a Customer object in Stripe. This is named `metadata` currently, so if other payment processors support metadata, it can be included in those. Currently, Stripe is the only one that implements this afaik.

```ruby
class User < ApplicationRecord
  pay_customer metadata: :stripe_metadata
  # pay_customer metadata: ->(user) { { user_id: user.id } }

  def stripe_metadata
    { user_id: id }
  end
end
```

Closes #431

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ### Unreleased
 
+* `pay_customer` now supports a metadata option to save on Stripe::Customers - @excid3
+
 # 3.0.24
 
 * Make payment method and charges consistent for Fake procsesor - @excid3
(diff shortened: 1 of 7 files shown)


excid3:
We can inject the `pay_customer_id` into the metadata automatically. This would be helpful since it's not easily accessible from the User model (in this example). We could pass it in as an argument as well and leave it up to the user.

excid3:
We're going to update the metadata each time the `email` field changes as well. Stripe will persist the metadata keys and merge them on update. The only way you'd lose a key/value is by sending `nil` as the value.

PR #587, opened 2022-03-31
update mailer to pull ActionMailer default from email address

arelenglish:
If the pay support_email isn't set, use the applications ActionMailer default from email address instead.


excid3:
Kept forgetting to add this, thanks a bunch!

arelenglish:
No problem! Happy to contribute!

PR #604, opened 2022-04-18
Add stripe receipt url to charge

mguidetti:
This adds the receipt url provided by the Stripe API Charge object to Pay's Charge model in a similar fashion to the existing `paddle_receipt_url`


excid3:
I added a test to confirm the `stripe_receipt_url` is being set.

mguidetti:
Ah thank you, I missed that.

PR #610, opened 2022-04-25
Allow custom mailer class

le-doude:
# Goal

Allow users of the gem to implement and inject their own Mailer for Pay.

# Delivery

This is inspired by how Devise allow you to specify your own mailer class.
Major difference is that we do not have access to newer version of `ActiveSupport::Dependencies` which have `constantize` and `references` methods helping make this more robust.
We fallback to using `ActiveSupport::Inflector#safe_constantize` instead.


le-doude:
@excid3 Hi, first PR on this project.
What other actions would you like me to take for this?

excid3:
@le-doude thanks for this!

Devise's works by setting the parent mailer class, so maybe we should go that route too. This seems a bit overkill, because you can copy the controller into your app/mailers folder if you want to override any of this. Maybe we can simplify and have the same `parent_mailer` config?

le-doude:
@excid3 thanks for your reply.

> Devise's works by setting the parent mailer class, so maybe we should go that route too. 

Devise gives you two points where you can customize the mailer:
1. `Devise::Mailer` is a subclass of your own `ApplicationMailer` defined in your own app by default. You can thus make sure it shares some base behaviour with all the mailers in your app, like layout, base path for templates or queues.
I believe this would be a good idea for Pay to do the same as this is almost a convension rails extension should all follow (IMHO). I did not need this for my use case though.
2. You can also customize the mailer by having your own implementation (that might or might not subclass `Devise::Mailer`) and load it using:
```ruby
# in initializer/devise.rb
Devise.setup do |config|
  config.mailer = 'MyCustomMailer'
end
```
This allows us to customize behaviour of the mailer for each action's context. Which is what I needed here and would not be covered by the first point or by just customizing the mail templates.

This PR and Devise use a string because of the difficulty of figuring loading order of classes. And that plus the fact that the webhooks call `Pay::UserMailer` directly made it hard to just override that class in a clean way. The string allows us to lazy load the mailer class once the whole app code has been loaded.

> This seems a bit overkill, because you can copy the controller into your app/mailers folder if you want to override any of this.

I am sorry I am not understanding what you propose here. 🙇 Can you give example?

> Maybe we can simplify and have the same parent_mailer config and maybe a generator to copy over the mailer views / controller to your app?

This would not have been sufficient for my use case but we can add it on top of this? In the end this approach I take can cover whatever customization necessary for the pay mailer.

As a final note: this approach also allows me to have very simple test for my custom mailer as it is just a normal mailer class in my app.

Thanks for reading

excid3:
This is missing a `parent_mailer` config to handle the subclass option like Devise has:

```ruby
  class Devise::Mailer < Devise.parent_mailer.constantize
```

We would need to add a `Pay.parent_mailer`

le-doude:
@excid3 I have added your suggestion. I'd like to write a test for it (maybe using the `dummy` app in test folder). Any suggestions?

PR #642, opened 2022-06-19
Use email_address_with_name helper in user_mailer

mguidetti:
This change adds the [ActionMailer::Base#email_address_with_name](https://github.com/rails/rails/pull/36454) helper added in Rails 6 to `UserMailer` so that customer names with special characters are escaped. 

Previously, the mailer could throw a `Net::SMTPSyntaxError` error if customer names had special characters. I ran across this with a customer who had a comma and period in their name. (eg: "Business Name, Inc.")


mguidetti:
Ah, my mistake. I thought this helper was added in 6.0, but looks like it was a 6.1 addition. Will look for a 6.0 compatible solution

mguidetti:
I extracted out rails 6.1's `email_address_with_name` helper here so that the gem can still be compatible with 6.0

excid3:
Thanks for this!

PR #645, opened 2022-06-26
Fix opts[:trial_from_plan] not respecting override

feliperaul:
When using `user.payment_processor.subscribe(plan: "xxx", trial_end: timestamp_in_the_future)`, I was getting the following error from Stripe:

> You cannot set `trial_end` or `trial_period_days` when `trial_from_plan=true`.

I then tried to pass `user.payment_processor.subscribe(plan: "xxx", trial_end: timestamp_in_the_future, trial_from_plan: false)`, but still kept getting the error.

Digging further, found a bug that `opts[:trial_from_plan]` will always be true even if you try to override it with false.

This PR makes so you can set `trial_from_plan` to false without it being overwritten to true.

### DIFF
--- a/lib/pay/stripe/billable.rb
+++ b/lib/pay/stripe/billable.rb
@@ -110,7 +110,7 @@ module Pay
         }.merge(options)
 
         # Inherit trial from plan unless trial override was specified
-        opts[:trial_from_plan] = true unless opts[:trial_period_days]
+        opts[:trial_from_plan] = opts.fetch(:trial_from_plan, true)
 
         # Load the Stripe customer to verify it exists and update payment method if needed
         opts[:customer] = customer.id


excid3:
I've tried to standardize `trial_period_days` as the primary way to set a trial so that you can use that across Braintree, etc. That's why it was using that here. If you're to pass the same arguments across to Braintree, you might run into invalid params. Maybe there's no good way around that. 

It might be better if we set `trial_from_plan` true only if there is no `trial_period_days` or `trial_end`. Then it can be seamlessly handled.

excid3:
I'm actually thinking that we should remove this. Stripe has deprecated the trial_period_days on plans, and now recommend passing `trial_period_days` or `trial_end` and this was just for some backwards compatibility.

A default trial is not good because a user could cancel and get infinite free trials, so that's why they deprecated this. It's probably wise to follow suit and make this a breaking change.

PR #647, opened 2022-06-26
Fixes non deterministic customer#subscription

feliperaul:
This PR fixes a non-deterministic `pay_customer#subscription` return if using Postgres.

That happened because in Postgres, if an order clause is not specified, the results return in non-deterministic order (https://stackoverflow.com/questions/6585574/postgres-default-sort-by-id-worldship). 

Psql will give the impression of returning records in ascending primary key (ID) order, but it turns out if you update a previously created record, it will start appearing first, before of the records with higher primary key. 

The bug only happened if `pay_customer#subscriptions` were already loaded; if they weren't, it would use `subscriptions.last`, which includes a sort clause in the query, and the bug didn't manifest; however, if they were loaded, for performance reasons, the gem used `subscriptions.reverse.detect {}`, and those already-loaded subscriptions were loaded in non-deterministic order.

The PR adds a default sort clause in the `has_many :subscriptions` association. 

I managed to write a failing test for psql; if you remove the `-> { order({ id: :asc }) }` this PR adds, that test will fail. 

This bug was causing random `pay_customer#subscription` returns depending on webhooks race condition (depending on the time of the subscription update, we'd get a different result as `pay_customer#subscription`. This PR fixes it.

### DIFF
--- a/app/models/pay/customer.rb
+++ b/app/models/pay/customer.rb
@@ -2,7 +2,7 @@ module Pay
   class Customer < Pay::ApplicationRecord
     belongs_to :owner, polymorphic: true
     has_many :charges, dependent: :destroy
-    has_many :subscriptions, dependent: :destroy
+    has_many :subscriptions, -> { order({ id: :asc }) }, dependent: :destroy
     has_many :payment_methods, dependent: :destroy
     has_one :default_payment_method, -> { where(default: true) }, class_name: "Pay::PaymentMethod"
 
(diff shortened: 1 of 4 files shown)


excid3:
Really great, thank you @feliperaul 👍 

I ended up just adding the sort to the `subscription` method. It makes it a little easier for overriding if someone wants to sort by a different column like `created_at` (which they might want to do if they're using UUIDs). I left it as sorting by ID because that should be fine and users don't need to add an index to created_at.

PR #648, opened 2022-06-27
Provides clearer guidance on how to upgrade to v4. 

evolve2k:
I was getting very lost and confused as to how much work was involved to upgrade to v4.

This PR attempts to make the documentation much simpler in terms of what people need to do to upgrade to v4.
I appreciate the core team may have their own preferences on formatting, hope that the content is acceptable to merge, feel free to adjust the headings/formatting to your preferred style.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,18 +16,53 @@ Pay.setup do |config|
 
 ### 4.0.0
 
-* [Breaking] Replaced `subscription` and `charge` email params to `pay_subscription` and `pay_charge` respectively. - @cjilbert504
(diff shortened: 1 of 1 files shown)


cjilbert504:
@evolve2k - I've pulled in a small example from this PR into the [Pay v3 -> v4 upgrade guide](https://github.com/pay-rails/pay/blob/master/UPGRADE.md#pay-30-to-40). Most of the information listed in this PR existed in that section of the upgrade guide already so that is why I only pulled a small example from this PR. We will try to make it clearer in the future to view the actual UPGRADE.md file in addition to reading the changelog when performing future upgrades. Thank you for taking the time to contribute this PR.

PR #661, opened 2022-07-14
Subscriptions pause

### LINKED ISSUE #659: Paused subscriptions from Stripe Portal are still marked as active
I have a few customers who've paused their subscription via Stripe Portal. Here's an example of one that I see in Stripe.

image

However, when I find this user's subscription in the console, Pay says that it is active.

```ruby
=> #<Pay::Subscription:0x000000010e06d2e0
 id: XX,
 customer_id: XXX,
 name: "XXXXXXXXXX",
 processor_id: "sub_XXXXXXXXXXXXXXX",
 processor_plan: "price_XXXXXXXXXXXXXXXXXX",
 quantity: 1,
 status: "active",
 trial_ends_at: nil,
 ends_at: nil,
 application_fee_percent: nil,
 metadata: {"pay_name"=>"full_time"},
 data: {},
 created_at: Tue, 12 Apr 2022 03:19:41.307379000 UTC +00:00,
 updated_at: Tue, 12 Apr 2022 03:19:41.307379000 UTC +00:00,
 prorate: true>
```

I'm not doing anything special with the webhooks but maybe I haven't set something up correctly? Is this behavior expected?

cjilbert504:
Resolves #659 - This PR addresses the issue of paused subscriptions still showing `active` with Stripe. It adds a new scope `not_paused` which filters records based on the `pause_behavior` accessor. It also adds an instance method that combines the `active?` and `paused?` instance methods for checking a given `Pay::Subscription` instance.


app/models/pay/subscription.rb
```diff
@@ -111,6 +113,10 @@ def active?
       ["trialing", "active"].include?(status) && (ends_at.nil? || on_grace_period? || on_trial?)
     end
 
+    def active_and_paused?
```
joemasilotti, line 116:
```suggestion
    def active_but_paused?
```

That might give a better idea as to why it's "weird" to be checking both.

excid3, line 116: A paused subscription is always considered active, so this seems redundant?

app/models/pay/subscription.rb
```diff
@@ -12,6 +12,8 @@ class Subscription < Pay::ApplicationRecord
     scope :cancelled, -> { where.not(ends_at: nil) }
     scope :on_grace_period, -> { cancelled.where("#{table_name}.ends_at > ?", Time.zone.now) }
     scope :active, -> { where(status: ["trialing", "active"], ends_at: nil).or(on_grace_period).or(on_trial) }
+    scope :not_paused, -> { where("data->>'pause_behavior' IS NULL") }
```
excid3, line 15: This scope probably needs to build upon the `active` scope since `paused` subscriptions are active ones.

```ruby
    scope :not_paused, -> { active.where("data->>'pause_behavior' IS NULL") }
```

`active_without_paused` maybe?

app/models/pay/subscription.rb
cjilbert504: I like it!

PR #668, opened 2022-07-26
Ensure Pay::Customer exists when client_reference_id is passed

excid3:
Stripe Checkout Sessions can include a `client_reference_id` to relate it to a record in your application. If this is set as a Signed GlobalID, Pay will now look up the record and make sure that it has a Stripe Pay::Customer record attached.

This is useful for using the new Stripe Pricing Table, or just with any Checkout Session:

```ruby
::Stripe::Checkout::Session.create(
  mode: "payment",
  client_reference_id: current_user.to_sgid,
  ...
)
```

The Pay webhooks will automatically lookup the record using it's Signed GlobalID and ensure they have a Pay::Customer record for Stripe.

cc @cjavdev

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 ### Unreleased
 
+* Support `client_reference_id` on Stripe Checkout Sessions - @excid3 @cjilbert504
+  This is helpful when using the Stripe Pricing Table or any Checkout Session. Requires a Signed GlobalID as the value to prevent tampering.
+
+```ruby
+::Stripe::Checkout::Session.create(
+  mode: "payment",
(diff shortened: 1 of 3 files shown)


cjavilla-stripe:
The only edge case I can think of is someone wants to use the client reference id for something other than the current user. 

In locate_owner should we check to see if the user was located before adding the processor?

excid3:
> The only edge case I can think of is someone wants to use the client reference id for something other than the current user.

You can pass in the SGID for any user. The current_user will probably be the most common, so I used that in the example, but you could replace it with any SGID.

> In locate_owner should we check to see if the user was located before adding the processor?

Good catch. Added tests for that just now. 👍

PR #677, opened 2022-08-12
Update installation doc to v5

bvalentino:
This **very minor** 😅 PR updates the version in the installation docs to v5.0, otherwise it fails when using v4.0 + Stripe v7:

``` sh
$ bin/rails pay:install:migrations
rails aborted!
[Pay] stripe gem must be version ~> 6
.../config/environment.rb:7:in `<main>'
Tasks: TOP => railties:install:migrations => db:load_config => environment
(See full trace by running task with --trace)
```


excid3:
Good catch! Thanks @bvalentino !

Issue #681, opened 2022-08-24
Send dunning email when stripe.invoice.payment_failed webhook is received


excid3:
We should add emails for failed payments to automatically notify customers to update their card.

This will use the `invoice.payment_failed` webhook from Stripe. It should be configurable in the Pay config.

Same should be done for Paddle and Braintree.

PR #684, opened 2022-08-25
WIP: configure email to be sent upon stripe payment failed webhook event

cjilbert504:
The groundwork for completing #681


app/views/pay/user_mailer/payment_failed.html.erb
```diff
@@ -0,0 +1,7 @@
+<h3>Processing your payment failed</h3>
+<p>
+  Your <%= Pay.business_name %> subscription payment failed. 
+  Please update your payment information.
+</p>
+
+<p>- The <%= Pay.business_name %> Team</p>
```
excid3, line 7: Some inspiration for content. This is one I send out currently, and they get more urgent after a couple attempts.

> Hey Collin,
> 
> Chris from GoRails here. I am writing because it looks like your subscription payment of $19 didn't go through. > Please [update your information here](https://gorails.com) and we'll give it another try.
> 
> Let us know if you have any questions,
> 
> —
> Chris
> Founder, GoRails

We can tweak this and include a link to the `root_url` for the user to click. We could try to make it feel friendly and use "The [business-name] team" or we could add another config for a name and include that (if present) to make it more human.

cjilbert504: Thanks for the thoughts here, I appreciate it. I was going to get with you about writing a nice email for this. This was just my placeholder/starting point for the beginning. :)

I'll cook up a updated one inspired by the comments here, and perhaps look at other dunning email examples elsewhere.

lib/pay/stripe/webhooks/payment_failed.rb
```diff
@@ -0,0 +1,23 @@
+module Pay
+  module Stripe
+    module Webhooks
+      class PaymentFailed
+        def call(event)
+          # Event is of type "invoice" see:
+          # https://stripe.com/docs/api/invoices/object
+
+          object = event.data.object
```
excid3, line 9: One thing we need to consider here:

The `attempt_count` field on the invoice shows the number of attempts to pay this invoice. I read somewhere in the docs that the first one can be from the initial attempt (like if someone is subscribing and their first attempt fails maybe?). That might need to skip the first attempt or something, but wanted to make a note about it.

That will also be useful for changing the text for these templates. Maybe the `payment_failed` mailer takes the attempt count and determines which template to render? Or we have separate `payment_failed_first`, `payment_failed_second`, etc mailers?

PR #687, opened 2022-08-31
Refactor pause behavior for Stripe and Paddle

excid3:
This standardizes some of the ways we handle pausing. Several of these are **breaking** changes to keep things more consistent.

1. Moves `paddle_paused_from` to `pause_starts_at` column
2. Removes `paused` from `active` scope
3. Adds `active_or_paused` scope
4. Removes `active_without_paused` scope
5. Adds `void` pause behavior grace period for Stripe
6. Adds `pause_behavior:string` column


joemasilotti:
Thanks for diving in so quickly, Chris! I know this isn't final yet, but I'm a bit confused as to what the new scopes mean.

Here are the four states a customer can be in on RailsDevs. Which scope and `Pay::Customer` method should I use to find each?

1. Subscription ends tomorrow - access granted
2. Subscription ended yesterday - access denied
3. Subscription ends tomorrow and is paused for future payment - access granted
4. Subscription ended yesterday and is paused for future payment - access denied

Ideally I don't need to know/care when a subscription is paused – I only need to know "does this user have access to paid content right now?".

excid3:
`active` is the scope of subscriptions that are considered active. A pause that starts in the future is considered on its grace period and therefore active.

You will want `active_or_paused` for rendering your billing page because you will need to show paused subscriptions so the user can resume.

excid3:
What's the confusion about?

joemasilotti:
Doesn't Stripe set a subscription to `active` even if it is paused? How do I know someone has paused their subscription but their paid ("grace period") has ended?

excid3:
Check the code for the scopes. We are handling that.

joemasilotti:
OK! After running the migration then manually syncing each subscription I think the scopes are working as I would expect. (`.active` is all I care about to show paid features.)

But it looks like `Pay::Customer#active?` isn't taking into account paused subscriptions, yet. Is that right or am I still missing something?

excid3:
You're gonna have to provide some examples / test cases so I can see what you're talking about.

joemasilotti:
Sorry Chris, I was using `#active?` when I should have been using `#subscribed?`.

I think this is all working as expected for me! Thanks so much for putting this together so quickly.

joemasilotti:
Hey Chris, any idea on when this will be merged and released? Is there any testing or debugging I can help with to get it over the finish line?

PR #694, opened 2022-09-10
Add current_period_start and current_period_end to Stripe subscriptions

mpvosseller:
This adds support for storing and reading `current_period_start` and  `current_period_end` on Stripe subscriptions. This is based on discussion https://github.com/pay-rails/pay/discussions/693.

Hoping to get feedback on this initial implementation.

- How do this Stripe implementation look?
- Is this how you would expect to handle the upgrade / db migration (i.e. just update `1_create_pay_tables.rb` directly and include a migration script in the upgrade documentation?)
- Tests sufficient?
- Would you want to add support for this in `fake_processor`?
- I'm not familiar with Braintree or Paddle but can look into it if you think it is required and no one else is available to.

Thanks!

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,24 @@
 
 Follow this guide to upgrade older Pay versions. These may require database migrations and code changes.
 
+## **Pay 5.0 to 6.0**
+This version adds support for accessing the start and end of the current billing period of a subscription. This currently only works with Stripe subscriptions.
+- Adds `current_period_start` and `current_period_end` to `Pay::Subscription`
+
+To upgrade you must add and run the following database migration.
+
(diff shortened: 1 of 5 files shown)


excid3:
This looks good @mpvosseller 👍 

Braintree has `billing_period_start_date` and `billing_period_end_date` that we can use.
https://developer.paypal.com/braintree/docs/reference/response/subscription

Paddle has `last_payment.date` and `next_payment.date` which are probably similar enough.
https://developer.paddle.com/api-reference/e33e0a714a05d-list-users

I think we're good on just including a migration in the docs. I'll probably be releasing another major version soon, so I plan to integrate this if we can.

PR #712, opened 2022-10-11
Closes #695 - create new stripe webhook for trial ending

### LINKED ISSUE #695: Add email for `stripe.customer.subscription.trial_will_end`
It would be great for Pay to send emails to remind customers when trials are ending.

> **customer.subscription.trial_will_end**
> Occurs three days before a subscription’s trial period is scheduled to end, or when a trial is ended immediately (using trial_end=now).

We might want two separate email contents for this so that we can denote the trial has ended or will end in X days.

cjilbert504:
Stripe will fire a `trial_will_end` event when a trial subscription is either 3 days from ending or ended immediately. This PR adds a new webhook that listens for that event and then routes to the correct mailer depending on the trial status e.g. is it ending in a few days or has it ended already?


lib/pay.rb
```diff
@@ -61,6 +61,7 @@ module Billable
   @@emails.subscription_renewing = ->(pay_subscription, price) {
     (price&.type == "recurring") && (price.recurring&.interval == "year")
   }
+  @@emails.subscription_trial_will_end = true
```
excid3, line 64: Do we need another setting for `subscription_trial_ended`?

app/mailers/pay/user_mailer.rb
```diff
@@ -20,6 +20,14 @@ def payment_action_required
       mail to: to
     end
 
+    def subscription_trial_will_end
+      mail to: to
+    end
+
+    def subscription_trial_ended
+      mail to: to
```
excid3, line 28: If you get a chance, merge master into this branch and swap this to use the new `mail_arguments`

lib/pay.rb
cjilbert504: Oh yeah good call, probably so.

node_modules/.yarn-integrity
```diff
@@ -0,0 +1,10 @@
+{
+  "systemParams": "darwin-x64-93",
+  "modulesFolders": [],
+  "flags": [],
+  "linkedModules": [],
+  "topLevelPatterns": [],
+  "lockfileEntries": {},
+  "files": [],
+  "artifacts": {}
+}
```
excid3, line 10: You might want to remove this from the commit. I don't think you meant it to be in here.

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
```
excid3, line 2: Same with this, I don't think you meant to include the yarn.lock. 👍

PR #726, opened 2022-11-10
added `from` method to Pay::ApplicationMailer

alexfilatov:
This `from` method will format the from email address in a way "Business Name <info@business.com>"

### DIFF
--- a/app/mailers/pay/application_mailer.rb
+++ b/app/mailers/pay/application_mailer.rb
@@ -1,14 +1,14 @@
 module Pay
   class ApplicationMailer < ActionMailer::Base
-    default from: from(Pay.business_name, Pay.support_email) || ApplicationMailer.default_params[:from]
-    layout "mailer"
+    def from_address(name, email)
+      return if email.blank?
 
-    def from(name, email)
-      return if email.blank? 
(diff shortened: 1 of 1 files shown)


excid3:
Hey @alexfilatov, you should be able to set `Pay.support_email = "Business Name <info@business.com>"` currently, so I don't think we need any changes here. That's the standard format for adding a display name to an email address.

alexfilatov:
yeah, I agree we can do that, but in this way, we are having a duplication of a business name in the config, and having the possibility to add a display name in the `support_email` is not that obvious. 
Anyway, thanks a lot for such a quick response, I'll close the PR draft. 

Also, what do you think, maybe it's worth updating docs with smth like that in the configuration part:

```
Pay.setup do |config|
 ...
  config.support_email = Address.new('Mikel Lindsaar (My email address) <mikel@test.lindsaar.net>')
 ...
end
```

alexfilatov:
on other hand, business name is not always a contact name (name of the manager in the business)

excid3:
I updated the docs to show how to set a name in the email. I think that will be good enough to show you how to set it up.

PR #743, opened 2022-11-29
Fixes localization of the default refund description

mpvosseller:
The default refund description was using the wrong key (`refund` instead of `pay.refund`). That caused the default refund description to be `translation missing: en.refund` and was viewable in Stripe's refund emails (and probably other places).


excid3:
Doh, thanks!

PR #744, opened 2022-11-30
Add clarity for flow of syncing attributes

cjilbert504:
Previously, it read a bit vaguely as to which direction customer attributes are syncing. This update hopefully provides a clearer description of which direction the sync flows.


excid3:
Ace! Thanks for the documentation PRs. 💖

PR #748, opened 2022-12-01
Fix NameError

pooriar:
Should declare class before referencing it

Co-authored-by: kaspth hey@kaspth.com


excid3:
Good catch, thanks @pooriar @kaspth!

kaspth:
@pooriar's working on a site that uses Pay through Jumpstart, so we paired on connecting the charge succeeded Webhook to a customer, ala:

```ruby
def call(event)
  Pay::Customer.find_by(processor_id: event.data.object.customer)
  # Find and connect the Webhook…
end
```

But yeah, then we found this 😄

PR #760, opened 2022-12-19
Add braintree subscription sync method

excid3:
This adds a `sync` method to the Braintree Subscription class.

Also changes Braintree subscriptions webhooks to sync instead of immediately operating on the subscription. This should make it more reliable in case webhooks arrive out of order, etc.


cjilbert504 (approved):
This is awesome! Love how reusable the `sync` method is in all of the webhooks!

PR #764, opened 2023-01-12
Adds Ruby 3.2 to the CI matrix.

petergoldstein:


excid3:
The nokogiri release was why I previously reverted Ruby 3.2 in CI. 

Looks like they just released 1.14 final so we can update that.

excid3:
sqlite3 is failing now

PR #771, opened 2023-01-30
add scope resolution operator to calls to Mail::Address

### LINKED ISSUE #770: NameError: uninitialized constant Pay::Mail after upgrading to 6.3.1
During my Ruby 3.2 upgrade took the time to also update all other gems, so pay got updated 
-  pay (6.1.0)
+ pay (6.3.1)

-  receipts (2.1.0)
+ receipts (2.2.0)

If I comment on the line where I set the support email in the pay initializer I get the app to load
`#  config.support_email = "helpme@example.com"`

otherwise, I cannot get zeitwerk to load the app:

```
rails zeitwerk:check 
rails aborted!
NameError: uninitialized constant Pay::Mail
```

cjilbert504:
This PR adds resolution to `Mail::Address` as a top level constant to avoid any confusion about this constant belonging under the `Pay` namespace.

Closes #770

### DIFF
--- a/lib/pay.rb
+++ b/lib/pay.rb
@@ -95,8 +95,8 @@ module Pay
   # Should return String or Array of email recipients
   mattr_accessor :mail_to
   @@mail_to = -> {
-    if ActionMailer::Base.respond_to?(:email_address_with_name)
-      ActionMailer::Base.email_address_with_name(params[:pay_customer].email, params[:pay_customer].customer_name)
+    if ::ActionMailer::Base.respond_to?(:email_address_with_name)
+      ::ActionMailer::Base.email_address_with_name(params[:pay_customer].email, params[:pay_customer].customer_name)
     else
       ::Mail::Address.new.tap do |builder|
         builder.address = params[:pay_customer].email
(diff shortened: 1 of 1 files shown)


lib/pay.rb
```diff
@@ -98,7 +98,7 @@ def self.mailer
     if ActionMailer::Base.respond_to?(:email_address_with_name)
       ActionMailer::Base.email_address_with_name(params[:pay_customer].email, params[:pay_customer].customer_name)
```
excid3, line 99: Should we also do this for `ActionMailer`?

cjilbert504: We probably should for consistency/clarity if nothing else.

lib/pay.rb
```diff
@@ -3,6 +3,7 @@
 require "pay/errors"
 require "pay/adapter"
 
+require "action_mailer"
 require "active_support/dependencies"
```
excid3, line 7: Not related to this PR exactly, but I wonder if we even need this?

Want to try removing it and see if all the tests pass?

cjilbert504: Yeah, that sounds good. I'll update now.

excid3:
This won't prevent every issue with users whole have customize the Rails loading, but it will help point them in the right direction.

Issue #562, opened 2022-03-09
Is it possible to set the plan name when using Stripe Checkout?


joemasilotti:
When using Stripe Checkout all of my subscriptions have the default name set in the Pay config. Now that I'm introducing new plans I'd like to differentiate them. I can backfill the existing ones without issue, but is it possible to set the plan name via Stripe Checkout?

If it helps, I'm not implementing the webhook but letting Pay handle it entirely.

excid3:
You can't because Stripe creates the subscription on their end.

Could probably add a `pay_name` metadata attribute and use that on `sync` instead of the default name.

joemasilotti:
I see the `enchantment` label, is it safe to assume I can't accomplish this unless Pay is changed or I implement my own webhooks?

joemasilotti:
Wow, that was fast! I'm a bit confused on how to set this up, though.

I set pay_name in the Stripe product for meta data and when creating the Stripe checkout link. Neither seem to apply. I am pointing to commit `4f4a726051bed3f88f6f74541cc2cdd8b600afb0`.

<img width="380" alt="image" src="https://user-images.githubusercontent.com/2092156/157581907-19819b58-25d6-4e47-ab9f-84f0227836cd.png">

```
def checkout
  user.set_payment_processor(:stripe)
  user.payment_processor.checkout(
    mode: "subscription",
    line_items: plan.price_id,
    success_url: analytics_event_url(event),
    metadata: {
      pay_name: "Foo Bar"
    }
  )
end
```

joemasilotti:
Quick update, I set the metadata on the _product_. If I set it via the Stripe dashboard I see it sync over to my local `Pay::Subscription`.

Maybe I'm missing something, but how can I set the metadata of a subscription before it exists? Shouldn't I be able to set it on the product and have it applied to the subscriptions?

excid3:
It should go on the `Subscription` metadata in Stripe.

excid3:
In your checkout example, you set the metadata on the Session, not the Subscription. You'll probably need to look at the Stripe API and see if you can set it on the Subscription.

joemasilotti:
Got it! Thanks for the help Chris, and super quick code change. I really appreciate it.

This is how I set the custom metadata when creating the Checkout URL:

```
def checkout
  user.payment_processor.checkout(
    mode: "subscription",
    line_items: plan.price_id,
    success_url: analytics_event_url(event),
    subscription_data: {
      metadata: {
[...]

PR #787, opened 2023-03-05
Update Stripe checkout docs

vasaf:
`pay_name` override should be inside metadata. See https://github.com/pay-rails/pay/issues/562#issuecomment-1064142965


excid3:
Thank you!

PR #792, opened 2023-03-15
Fix example link in Paddle passthrough example

RyanJamesCaldwell:
This PR updates the Paddle checkout example in the documentation to produce a valid hyperlink - the current example doesn't have a closing `</a>` tag or any text.


excid3:
Thanks @RyanJamesCaldwell 👍

PR #815, opened 2023-05-13
Lazy instantiate webhook callback classes

t27duck:
Rails doesn't load (or can access) anything constants in the app directory during app boot. To have custom webhook callbacks, the constants have to be defined in the initializer or manually required from lib.

This change allows `Pay::Webhooks.delegator.subscribe` to take a string argument. When `NotificationAdapter` is invoked and the subscriber is a string, we'll `constantize` and instantiate it before calling it.

### DIFF
--- a/test/pay/webhooks/delegator_test.rb
+++ b/test/pay/webhooks/delegator_test.rb
@@ -5,6 +5,7 @@ class Pay::WebhookDelegatorTest < ActiveSupport::TestCase
     attr_accessor :success
 
     def call(event)
+      event[:result] = true
       @success = true
     end
   end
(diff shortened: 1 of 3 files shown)


test/pay/webhooks/delegator_test.rb
```diff
@@ -40,6 +41,13 @@ def call(event)
     assert processor.success
   end
 
+  test "can subscribe with a string" do
```
t27duck, line 44: Can't say I'm thrilled with this test. Basically taking advantage of Ruby's pass-by-ref and relying on the event object being mutated...

This is also hard to replicate the Rails boot process as the processor class here has already been defined. Going to assume `NotificationAdapter` isn't actually invoked to run the processor until app boot is done (which should be the case?).

t27duck:
Closing in favor of doc update showing how to use Rails' initializer hooks.

Issue #789, opened 2023-03-09
Sync Stripe Credit Balance to Pay::Customer


excid3:
If a user has an invoice credit balance, we should save this value on the Pay::Customer so that we can use this to preview upcoming invoices and the applied credit balance.

```ruby
customer = ::Stripe::Customer.retrieve(id: "cus_1234", expand: ["invoice_credit_balance"])

#<Stripe::Customer:0x2c2e0 id=cus_4QFOF3xrvBT2nU> JSON: {
  "id": "cus_4QFOF3xrvBT2nU",
  "object": "customer",
  "address": {"city":"City9debc82d-10b2-42e2-87ec-3316c9e7acc5","country":"Country2bde78f1-d28b-4c80-8e99-836828661c29","line1":"Line136e8abc0-92e2-4b91-aa83-b07ee5305ef0","line2":"Line2139281af-fd20-4e7e-8a99-2a09c32dbe1e","postal_code":"PostalCode190108c3-177d-4b5c-9c5a-89ef946e36b5","state":"State8618eae4-ad11-4577-b227-ae6667c2e41e"},
  "balance": 0,
  "created": 1405641986,
  "currency": "usd",
  "default_source": "card_14HOtJ2eZvKYlo2CD2lt4r4W",
  "delinquent": false,
  "description": "someone@example.com for Coderwall",
  "discount": null,
  "email": "namee97851f5-1cf1-48f6-b694-497e641661e8@test.com",
  "invoice_credit_balance": {"usd":0},
  "invoice_prefix": "93EC0E1",
  "invoice_settings": {"custom_fields":null,"default_payment_method":"pm_1MWFdw2eZvKYlo2CvWcM1Btd","footer":null,"rendering_options":null},
  "livemode": false,
  "metadata": {"Address":"testadress","CustomerReferenceId":"16527612","CustomerReferenceType":"reedonline","address":"testadress","adress":"testadress","email":"test@test.com","meta-key":"meta-value","order_id":"6735","order_id2":"6735","order_id_1":"6735","tag-key":"tag-value"},
  "name": "namee97851f5-1cf1-48f6-b694-497e641661e8",
  "next_invoice_sequence": 962401,
  "phone": null,
  "preferred_locales": [],
  "shipping": null,
  "tax_exempt": "none",
  "test_clock": null
}
```

Looks like we'll have to grab the correct currency to retrieve the invoice_credit_balance.

```ruby
customer.invoice_credit_balance[customer.currency]
#=> 0
[...]

PR #825, opened 2023-06-06
feat: Sync Stripe Credit Balance to Pay::Customer

nachiket87:
fixes https://github.com/pay-rails/pay/issues/789

First time contributing to this repo. ❤️


excid3:
This is a great start @nachiket87 👍 

I think we'll want to store this in the Pay::Customer object and keep it synced using the [customer.updated webhook](https://stripe.com/docs/api/events/types#event_types-customer.updated). This way we can have it in the database always and keep it in sync with what's in Stripe (like when someone manually applies credit to a customer in their dashboard).

nachiket87:
> This is a great start @nachiket87 👍
> 
> I think we'll want to store this in the Pay::Customer object and keep it synced using the [customer.updated webhook](https://stripe.com/docs/api/events/types#event_types-customer.updated). This way we can have it in the database always and keep it in sync with what's in Stripe (like when someone manually applies credit to a customer in their dashboard).

ah yes I see now! I'll make the changes tomorrow. Thanks for the feedback!

nachiket87:
Hey @excid3 I've updated the PR with the new changes. If you get a minute, please could you take a look. Thanks!

nachiket87:
Thanks for the refactor @cjilbert504!

lib/pay/stripe/webhooks/customer_updated.rb
```diff
@@ -9,14 +9,21 @@ def call(event)
...
+          if (payment_method_id = stripe_customer.invoice_settings.default_payment_method)
             Pay::Stripe::PaymentMethod.sync(payment_method_id, stripe_account: event.try(:account))
 
           else
             # No default payment method set
             pay_customer.payment_methods.update_all(default: false)
           end
+
+          # Sync invoice credit balance and currency
+          if stripe_customer.invoice_credit_balance&.keys.present?
```
cjilbert504, line 24: @nachiket87 - I don't think we need to do this call to `.keys` here. It feels like we just want to check that the resulting object isn't empty. So I think we could just do `stripe_customer.invoice_credit_balance.present?` instead. I think it reads better additionally i.e. "Does the customer have an invoice credit balance?" versus "Does the customer have invoice credit balance keys?"

As an example:
```ruby
stripe_customer = OpenStruct.new(invoice_credit_balance: {"usd" => 12345})
=> #<OpenStruct invoice_credit_balance={"usd"=>12345}>
stripe_customer.invoice_credit_balance.present?
=> true

stripe_customer = OpenStruct.new(invoice_credit_balance: {})
=> #<OpenStruct invoice_credit_balance={}>
stripe_customer.invoice_credit_balance.present?
=> false
```

nachiket87: Thanks @cjilbert504 I've refactored it in 968f6767142c56558d0161ea0825886b6e1b95ea

excid3: I ended up saving the whole `invoice_credit_balance` just in case there's a need for someone to access the other currencies. This will be the safest / most future proof.

excid3:
Huge thanks @nachiket87! This is going to be really handy to have.

nachiket87:
Thank you @excid3 got taking the time to review and merge and @cjilbert504 for the reviews and suggestions

cjilbert504:
@nachiket87 - Thanks a bunch for this one! Nice work

PR #832, opened 2023-07-13
Adds documentation related to Stripe customer reconciliation (resolves #831)

### LINKED ISSUE #831: Stripe customer creation is performed without processor reconciliation for existing Stripe customers
I was working on a project in a development environment when I ran into an interesting situation where duplicate Stripe customers are created if the Rails application using pay-rails is not aware of existing Stripe customers already stored by Stripe with the same customer attributes (e.g. email). 

I encountered this by creating subscriptions and charges for a User within the project I'm developing, then resetting the project's Rails database, thus clearing all references to the Stripe processor values, and re-starting the application with the clean database. Attempting to create more subscriptions and charges on the same Stripe customer (e.g. Rails User with same email address) creates a new Stripe customer with the same email associated, resulting in a duplicate customer within Stripe.

I don't think that this is technically a bug, as I could see it being the Rails app's responsibility to reconcile this situation (such as calling `update_customer` with the appropriate `processor_id`), as no duplicates would occur if the application is aware of all created Stripe customers (which would be typical and expected for a production environment). However, I think if this is intended functionality, it might require more prominent documentation, and potentially a configuration option for processor customer reconciliation to make things "just work" in an opinionated way.

# Reproduction
Assuming the following `User` definition:

```ruby
class User < ApplicationRecord
  pay_customer default_payment_processor :stripe
end
```

When a `User` is created, the `payment_processor` holds null values related to the customer object:

```
#<Pay::Customer:0x0000562180830ed0
 id: 1,
 owner_type: "User",
 owner_id: 2,
 processor: "stripe",
 processor_id: nil,
 default: true,
 data: nil,
 deleted_at: nil,
 created_at: Tue, 11 Jul 2023 20:34:08.985920000 UTC +00:00,
 updated_at: Tue, 11 Jul 2023 20:34:08.985920000 UTC +00:00,
 plan: nil,
 quantity: nil,
 payment_method_token: nil>
```
Attempting to sync subscriptions using `cur_user.payment_processor.sync_subscriptions` results in the creation of a new Stripe customer, and the Stripe customer id is stored as the `processor_id`. The creation of the Stripe customer is done without attempting to identify if a Stripe customer already exists (within Stripe, not the Rails application) with the same email or other customer attributes, and as a result, two Stripe customers with the same attributes exist.

btonic:
* Provides details on how to manually reconcile existing Stripe customers with Pay's customer records.
* Provides examples on how to automatically reconcile Stripe customers with Pay's customer records using ActiveRecord callbacks.
* Describes how to backfill subscriptions and charges for reconciled customers.

### DIFF
--- a/docs/stripe/9_customer_reconciliation.md
+++ b/docs/stripe/9_customer_reconciliation.md
@@ -1,0 +1,65 @@
+# Stripe customer reconciliation
+Pay tracks customers for each payment processor using the `Pay::Customer` model, but the payment processor logic for customers varies between providers. When using Stripe with Pay, a customer object must exist for a model with `pay_customer` for charges and subscriptions to occur. If a `Pay::Customer` does not exist, one will be created automatically when attempting to operate upon subscriptions and charges.
+
+When creating the new `Pay::Customer`, Pay does not attempt to reconcile the attributes used to create a `Pay::Customer` with existing Stripe customers. As a result, there is a possibility that duplicate Stripe customers may exist with the same attributes (e.g. email) if the application using Pay does not manually reconcile existing Stripe customers with the `Pay::Customer` s.
+
+### Manual reconciliation
+The Stripe API can be used to list all existing Stripe customers. This allows the application to implement the necessary logic for creating and associating `Pay::Customer` s within the application.
+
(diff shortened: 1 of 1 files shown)


excid3:
This is excellent @btonic! Thank you so much!

PR #834, opened 2023-07-17
Removed references to Rails.applications.secrets

heliocola:
### Goal
Remove usage of deprecated Rails.applications.secrets

### Deprecation warning on CI
```
DEPRECATION WARNING: `Rails.application.secrets` is deprecated in favor of `Rails.application.credentials` and will be removed in Rails 7.2. (called from secrets at /Users/hac/.rvm/gems/ruby-3.2.2@railsdevs.com/gems/pay-6.6.1/lib/pay/env.rb:43)
```

### DIFF
--- a/lib/pay/env.rb
+++ b/lib/pay/env.rb
@@ -11,6 +11,7 @@ module Pay
     # 1. Check environment variable
     # 2. Check environment scoped credentials
     # 3. Check unscoped credentials
+    # 4. Check scoped and unscoped secrets (removed in Rails 7.2)
     #
     # For example, find_value_by_name("stripe", "private_key") will check the following in order until it finds a value:
     #
@@ -20,8 +21,8 @@ module Pay
     def find_value_by_name(scope, name)
       ENV["#{scope.upcase}_#{name.upcase}"] ||
         credentials&.dig(env, scope, name) ||
-        secrets&.dig(env, scope, name) ||
         credentials&.dig(scope, name) ||
+        secrets&.dig(env, scope, name) ||
         secrets&.dig(scope, name)
     rescue ActiveSupport::MessageEncryptor::InvalidMessage
       Rails.logger.error <<~MESSAGE
@@ -38,15 +39,11 @@ module Pay
     end
 
     def secrets
-      is_rails_version_below_7_2? ? Rails.application.secrets : nil
+      Rails.application.secrets if Rails.application.respond_to?(:secrets)
     end
 
     def credentials
       Rails.application.credentials if Rails.application.respond_to?(:credentials)
-    end
-
-    def is_rails_version_below_7_2?
-      Rails.gem_version < Gem::Version.new("7.2")
     end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/pay/env.rb
```diff
@@ -9,22 +9,18 @@ module Env
...
     #   Rails.application.credentials.dig(:production, :stripe, :private_key)
-    #   Rails.application.secrets.dig(:production, :stripe, :private_key)
     #   Rails.application.credentials.dig(:stripe, :private_key)
-    #   Rails.application.secrets.dig(:stripe, :private_key)
     def find_value_by_name(scope, name)
       ENV["#{scope.upcase}_#{name.upcase}"] ||
         credentials&.dig(env, scope, name) ||
-        secrets&.dig(env, scope, name) ||
-        credentials&.dig(scope, name) ||
-        secrets&.dig(scope, name)
```
excid3, line 27: We need to keep these for backwards compatibility. Instead of deleting them, we can check if we're running Rails 7.2 higher and skip secrets.

excid3 (changes requested):
Whoops, closed this on accident.

Let's include Rails secrets if the app is on 7.1 or lower. We don't want to break existing applications.

lib/pay/env.rb
heliocola: @excid3 : who do you like this to access and compare Rails version?
```
   Rails::version.gsub(".", "").to_i
```

that will turn "7.0.6" to 706 (integer) and I can do integer comparison.
I can also do:
```
  Rails::verstion.to_f
```

But, that transform "7.0.6" in 7.0 (float).
And it will transform "7.1.***" to 7.1 which can I can do the comparison.

Or do you have any other suggestion?

lib/pay/env.rb
```diff
@@ -40,11 +38,15 @@ def env
...
-      Rails.application.secrets
+      is_rails_newer_or_equal?(720) ? nil : Rails.application.secrets
     end
 
     def credentials
       Rails.application.credentials if Rails.application.respond_to?(:credentials)
     end
+
+    def is_rails_newer_or_equal?(version)
+      Rails.version.delete(".").to_i >= version
```
heliocola, line 49: @excid3 : after I've pushed this code I've realized that Rails version 7.1.2.1 will break this logic, so I changed the PR to `Draft` and will work on that later.

And also remembered this I saw this in the code, while looking around `Rails.gem_version >= Gem::Version.new("6.1")`.

heliocola:
After a few back and forth I think I've got to a point that this is ready for review! :-)

lib/pay/env.rb
```diff
@@ -40,11 +38,15 @@ def env
     end
 
     def secrets
-      Rails.application.secrets
+      is_rails_version_below_7_2? ? Rails.application.secrets : nil
```
excid3, line 41:
```suggestion
      Rails.application.secrets if Rails.application.respond_to?(:secrets)
```
Does this work instead? Then it's consistent with `credentials` and we don't need any code to check the specific Rails version.

excid3: I guess this would still print the deprecation warning in Rails 7.1, but it would handle Rails 7.2 safely.

excid3 (approved):

excid3:
Thanks very much @heliocola!! 👍

PR #836, opened 2023-08-09
Respect configured primary key type in migration

MSchmidt:
Instead of changing the migration manually this will respect the setting from the rails project. Implementation is the same as ActiveStorage uses.

### DIFF
--- a/docs/1_installation.md
+++ b/docs/1_installation.md
@@ -36,8 +36,6 @@ Copy the Pay migrations to your app:
 bin/rails pay:install:migrations
 ````
 
->If your models rely on non integer ids (uuids for example) you will need to alter the migrations.
-
 Then run the migrations:
 
 ```bash
@@ -115,12 +113,12 @@ If you want to automatically sync whenever any other attribute changes, override
 
 ```rb
 class User < ApplicationRecord
-  
+
   def pay_should_sync_customer?
     # super will invoke Pay's default (e-mail changed)
     super || self.saved_change_to_address? || self.saved_change_to_name?
   end
-  
+
 end
 ```
 


db/migrate/1_create_pay_tables.rb
```diff
@@ -12,8 +15,8 @@ def change
     add_index :pay_customers, [:owner_type, :owner_id, :deleted_at, :default], name: :pay_customer_owner_index
     add_index :pay_customers, [:processor, :processor_id], unique: true
 
-    create_table :pay_merchants do |t|
-      t.belongs_to :owner, polymorphic: true, index: false
+    create_table :pay_merchants, id: :uuid do |t|
```
excid3, line 18:
```suggestion
    create_table :pay_merchants, id: primary_key_type do |t|
```

These should use the `primary_key_type` variable, right?

excid3:
This is great @MSchmidt! Didn't realize ActiveStorage had this and had wanted a solution for a while like this.

MSchmidt:
Oh you are absolutely right. I adjusted this in my project and then messed the PR up. Will fix

db/migrate/1_create_pay_tables.rb
MSchmidt: 🙈 Fixed!

excid3:
Perfect. Thanks again for this. Super duper handy and nice to see that Rails uses this approach internally!

MSchmidt:
Thanks for merging so quickly. Wondering if this should be part of `ActiveRecord::Migration` then. Not very DRY if every project needs the same few lines of code.

excid3:
Sure seems like it could be extracted as a couple methods for easy access because any Rails engine with migrations _should_ use it. Maybe you could get it in before Rails 7.1?

Tag me on the PR if you make one to Rails!

PR #838, opened 2023-08-14
Added support for Paddle Billing

### LINKED ISSUE #837: Support for Paddle Billing
Paddle have been working on improving their service for SaaS companies and have launched [Paddle Billing](https://www.paddle.com/billing). I've updated my Ruby library called [PaddleRB](https://github.com/deanpcmad/paddlerb) and would like to update Pay to take advantage of the new APIs.

Paddle Classic, which the previous service has been renamed to, will continue to function so existing code will need to be kept.

My proposal is for there to be an additional payment provider, maybe called `paddle_billing`, which would be for the new service. This would allow previous subscriptions, etc to still work fine and any new customers, subscriptions, transactions will be separate.

Would you be happy for me to do this? And if so, anything to consider? I'll likely clone the existing paddle code and update as required.

deanpcmad:
This PR adds support for Paddle Billing, the new service/APIs available from Paddle. [More Info](https://www.paddle.com/billing)

- Old Paddle code has been renamed to Paddle Classic
- Existing subscriptions will need to have their processor changed from `paddle` to `paddle_classic`

Paddle Billing now supports multiple subscriptions for a customer, similar to Stripe, so I've copied some code from Stripe and modified as required.

This uses my [Paddle Gem](https://github.com/deanpcmad/paddle) which supports both the new Billing but also the Classic APIs. The existing Paddle Classic code is untouched, but could be changed to use my library.

Fixes #837


excid3:
Dang! I will try and find some time this week to review.

> The existing Paddle Classic code is untouched, but could be changed to use my library.

Let's do this in a second PR (before release). Consolidating dependencies would be nice and I'm sure that users on Paddle Classic will likely want to upgrade so they would only need the single dependency in the new version.

deanpcmad:
It's not complete yet, just thought I'd create a PR of what I've done so far :sweat_smile: 

> Let's do this in a second PR (before release). Consolidating dependencies would be nice and I'm sure that users on Paddle Classic will likely want to upgrade so they would only need the single dependency in the new version.

Sounds good :+1:

strzibny:
@deanpcmad It's great that you already started a PR.

I too looked into Paddle Billing recently. For one, I already documented the new webhook verification so you can reuse it for the PR: https://nts.strzibny.name/verify-paddle-billing-webhooks-rails/

I have one question. Should we design the pause & cancellations changes as they are in Paddle Billing?

In Paddle, subscriptions are active even on a grace period which looks like Pay now follows as well. Hovewer it seems to me that Pay still marks subscriptions as "paused" or "cancelled" even if the change is only scheduled.

What would be the right way for this PR to work?

I would maybe prefer to have this same as Paddle since it's less mental overhead. On the other hand I understand the value of doing it similarly as for other providers.

@excid3 can you provide any insight on the design here?

deanpcmad:
@strzibny thanks for the link! 

I'm not sure how the paused and cancelled subscriptions would work tbh. I see that they created a "scheduled change" with the `effective_at` field, which I'm currently using.

strzibny:
Yep, exactly as you said, they should be modeled with scheduled changes I believe. But previously Pay would immediately marked a scheduled subscription to be paused or cancelled. I would like to avoid that and match the status in Paddle, whatever it is.

deanpcmad:
Maybe setting the `pause_starts_at` field when a pause event happens and then when the webhook fires when the pause actually happens, then update the status to paused?

Also, when upgrading a subscription, Paddle creates a Credit, so when creating transactions that will need to be taken into account

wn5XzLOSQG

deanpcmad:
Not sure why the tests are failing at the mo

app/controllers/pay/webhooks/paddle_controller.rb
```diff
@@ -6,26 +6,37 @@ class PaddleController < Pay::ApplicationController
...
         Pay::Webhooks::ProcessJob.perform_later(record)
       end
 
-      def verified_event
-        event = verify_params.as_json
-        verifier = Pay::Paddle::Webhooks::SignatureVerifier.new(event)
-        return event if verifier.verify
-        raise Pay::Paddle::Error, "Unable to verify Paddle webhook event"
+      # Pass Paddle signature from request.headers["Paddle-Signature"]
+      def valid_signature?(paddle_signature)
```
excid3, line 27: Is verification the same between Classic and Billing?

deanpcmad:
I received an email about a comment on this but I can't see it now 🧐 
But no, Classic and Billing use different ways of verifying the signature. 
I've not had much chance to work on this, but I can give you access to my repo?

excid3:
Whoopsie. I did not mean to merge that lol. 🙃

deanpcmad:
Also need to work on tests for Paddle

excid3:
I see that PaddlePay is still referenced in the new Paddle code, so we should update all those references as well.

All we need for the scheduled changes is to use those in the `sync` method to determine the date, just like we do with Stripe. Should be easy enough.

excid3:
@deanpcmad I see the `custom_data.passthrough` is still referenced in Subscription#sync. That can be removed too?

deanpcmad:
Sounds good, so still set the status and date fields as required? I believe I already have that set when running `pause` or `cancel` on the subscription, so that can be added so the `sync` method too.

And yes, that can be removed as I found that it's much easier and cleaner to set the Customer ID when setting up the Paddle subscription button.

I can help work through the docs, as there's a few changes on their new Paddle.js

excid3:
Yeah, status should always be set to what Paddle has defined.

I'll go ahead and remove that. 👍 

Docs + tests are going to be the biggest things to complete I think.

excid3:
I think that `charge` needs either an update or an additional method since Paddle Billing introduced transactions (but they require items with a price ID instead of amounts).

excid3:
After upgrading Jumpstart Pro... I'm starting to think it's better if we use `Pay::PaddleBilling` and `Pay::PaddleClassic`. Much easier to differentiate and less confusing for existing users. They're likely to keep that naming in their site / docs for a long time. 

I will probably rename it today or tomorrow unless you beat me to it @deanpcmad.

deanpcmad:
Yeah that makes sense tbh.  Some tests need to be written for it too

excid3:
I'll start a PR for it today.

PR #841, opened 2023-08-18
Update documentation for reconciling stripe customer subscriptions

Willardgmoore:
Update documentation for reconciling stripe customer subscriptions

If you prefer to update the documentation, choose this PR. 
If you prefer to make the original method call available, [choose this other one](https://github.com/pay-rails/pay/pull/840).

### DIFF
--- a/docs/stripe/9_customer_reconciliation.md
+++ b/docs/stripe/9_customer_reconciliation.md
@@ -56,7 +56,7 @@ end
 ### Backfilling subscriptions and charges for reconciled customers
 When `Pay::Customer`s are created using `Model.set_payment_processor` or `Model.add_payment_processor`, existing Stripe subscriptions and charges are not automatically backfilled.
 
-To backfill active subscriptions and the charges associated with those subscriptions, the `Pay::Customer.sync_subscriptions` method can be used. To backfill all subscriptions including canceled subscriptions, the `status: "all"` parameter can be provided (e.g. `Pay::Customer.sync_subscriptions(status: "all")`).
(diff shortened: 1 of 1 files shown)


docs/stripe/9_customer_reconciliation.md
```diff
@@ -56,7 +56,7 @@ end
 ### Backfilling subscriptions and charges for reconciled customers
 When `Pay::Customer`s are created using `Model.set_payment_processor` or `Model.add_payment_processor`, existing Stripe subscriptions and charges are not automatically backfilled.
 
-To backfill active subscriptions and the charges associated with those subscriptions, the `Pay::Customer.sync_subscriptions` method can be used. To backfill all subscriptions including canceled subscriptions, the `status: "all"` parameter can be provided (e.g. `Pay::Customer.sync_subscriptions(status: "all")`).
+To backfill active subscriptions and the charges associated with those subscriptions, the `Pay::Stripe::Billable.new(pay_customer).sync_subscriptions` method can be used. To backfill all subscriptions including canceled subscriptions, the `status: "all"` parameter can be provided (e.g. `Pay::Stripe::Billable.new(pay_customer).sync_subscriptions(status: "all")`).
```
excid3, line 59: This should probably be `@user.payment_processor.sync_subscriptions` 

When I originally wrote it, I was trying to show that #sync_subscriptions is a method on the Pay::Customer object, but this is what the actual code would look like.

We don't want people reaching into Pay::Stripe::Billable because it isn't really an official interface and is subject to change.

PR #846, opened 2023-09-07
Updates Stripe gem version in the docs

BilalBudhani:
The docs still refer to v8 whereas the gem enforces v9

image


excid3:
Thanks @BilalBudhani!

PR #847, opened 2023-09-25
Implement Processor for Lago

moxvallix:
Lago is an open source billing management software. It doesn't handle payments itself, rather it manages subscriptions and invoicing, and hands off charges to other services such as stripe.

https://docs.getlago.com/guide/introduction/welcome-to-lago

This PR implements all the necessary classes/methods/webhooks for Pay to integrate (mostly) with Lago.

There is room for improvement, maybe would be nice for some helpers for setting up Lago with Stripe etc, but I believe that this should be enough for working with Lago for now.

Lago has numerous differences to other billing systems, and I have tried my best to fit Pay's way around it.

Here are a few notable things about Lago:

* Lago itself doesn't handle payments, but can be set up to trigger them through other services (Stripe etc.)
  - Currently Pay doesn't provide an interface for setting up payment providers in Lago.
  - Payment providers will have to be configured directly with the Lago API or UI.
* Lago subscriptions do not have trials etc.
* Charges are mapped to Lago invoices.
* Some features require [Lago Premium](https://www.getlago.com/pricing) to function correctly.
* Wherever Lago requires an external_id, Pay will use the [GlobalID](https://github.com/rails/globalid) of the corresponding Pay object.
  - This is unless the object already has processor_id set, in which case it'll use that.

I'm not the most familiar with writing tests, but I have tried my best with them. Hopefully they are good.

I have also fully written out documentation for using the Lago processor. It doesn't cover *every* method, but covers the main ones that should be used.


excid3:
This is great @moxvallix! 

We've had trouble maintaining payment processors that we personally don't use, so I'm thinking we might want to make this a separate gem that users could install alongside Pay to add Lago support.

moxvallix:
For anyone coming across this in the future, lago support is now in the pay-lago gem.

https://github.com/demingfactor/pay-lago

Issue #849, opened 2023-09-28
`create_pay_tables.pay.rb` migration file fails to run if `strong_migrations` gem installed


DaniG2k:
I just tried running the migrations from `create_pay_tables.pay.rb`. These fail to execute if the `strong_migrations` gem installed. The failure occurs with the following output:

```bash
StandardError: An error has occurred, this and all later migrations canceled:

=== Best practice #strong_migrations ===

Adding a non-unique index with more than three columns rarely improves performance.
Instead, start an index with columns that narrow down the results the most.
```

The failing line is:

```ruby
add_index :pay_customers, [:owner_type, :owner_id, :deleted_at, :default], name: :pay_customer_owner_index
```

but I suspect any `add_index` in the migration file that does not have a `unique: true` constraint will also fail to run with `strong_migrations`. Not really a bug I suppose, but the `pay` gem's migration file can probably use a bit of improvement in this respect.

excid3:
Could possibly make `[:owner_type, :owner_id, :default]` a unique index as long as nulls are handled properly across all databases.

PRs welcome

PR #850, opened 2023-09-30
Fix for `strong_migrations`

DaniG2k:
Closes https://github.com/pay-rails/pay/issues/849

This PR modifies the `pay_customer_owner_index` index to rely on 3 columns instead of 4, so that migrations pass for people using the `strong_migrations` gem. I've also made this index unique as per [Chris's comment](https://github.com/pay-rails/pay/issues/849#issuecomment-1739146919).

Here's the explanation as to why one should [keep non-unique indexes to three columns or less](https://github.com/ankane/strong_migrations#keeping-non-unique-indexes-to-three-columns-or-less)

### DIFF
--- a/db/migrate/1_create_pay_tables.rb
+++ b/db/migrate/1_create_pay_tables.rb
@@ -11,7 +11,7 @@ class CreatePayTables < ActiveRecord::Migration[6.0]
       t.datetime :deleted_at
       t.timestamps
     end
-    add_index :pay_customers, [:owner_type, :owner_id, :default], name: :pay_customer_owner_index, unique: true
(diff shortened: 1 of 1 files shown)


db/migrate/1_create_pay_tables.rb
```diff
@@ -11,7 +11,7 @@ def change
       t.datetime :deleted_at
       t.timestamps
     end
-    add_index :pay_customers, [:owner_type, :owner_id, :deleted_at, :default], name: :pay_customer_owner_index
```
excid3, line 14: Now that I'm thinking about it, `:deleted_at` might be the safer one to keep. `:default` is a boolean and in theory a faster query?

PR #851, opened 2023-10-05
Make sure onboarding_complete? returns a boolean value

olimart:
This PR ensures we return a proper boolean value when `Pay::Merchant.data` is nil.
Also updated the tests to describe the expected return value.


olimart:
Thanks Chris @excid3

PR #860, opened 2023-10-16
Canceled subscriptions are now always inactive

excid3:
This makes all `status: :canceled` subscriptions as inactive.

Previously, `ends_at` could be set to a future time and a `canceled` subscription would still be considered active until that time. This was implement to make Braintree consisted where canceling a subscription during a trial would have no way of providing access until the end of the trial.

We've decided to make Braintree end the trial immediately and treat all `canceled` subscriptions as inactive.

### DIFF
--- a/lib/pay/paddle/subscription.rb
+++ b/lib/pay/paddle/subscription.rb
@@ -82,7 +82,9 @@ module Pay
       end
 
       def cancel(**options)
-        if on_trial?
+        return if canceled?
+
+        ends_at = if on_trial?
           trial_ends_at
         elsif paused?
           pause_starts_at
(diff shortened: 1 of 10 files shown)


lib/pay/paddle/subscription.rb
```diff
@@ -82,7 +82,9 @@ def subscription
       end
 
       def cancel(**options)
-        if on_trial?
+        return if canceled?
+
+        ends_at = if on_trial?
```
excid3, line 87: Not entirely sure why we have this if statement... it doesn't do anything?

excid3: Looks like this done in the Paddle Billing update.

test/models/pay/subscription_test.rb
```diff
@@ -305,6 +305,24 @@ class Pay::Subscription::Test < ActiveSupport::TestCase
...
+    assert_not_includes @pay_customer.subscriptions.active, @subscription
+  end
+
+  test "unpaid" do
+    @subscription.update!(status: :unpaid)
+    refute @subscription.active?
+    assert_not_includes @pay_customer.subscriptions.active, @subscription
+  end
+
+  test "canceled subscriptions with a future ends_at are considered active" do
```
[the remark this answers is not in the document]
excid3:
```suggestion
  test "canceled subscriptions with a future ends_at are inactive" do
```

cjilbert504 (approved):

PR #873, opened 2023-10-18
Paddle Billing Docs

deanpcmad:
This PR updates the documentation to convert Paddle > Paddle Classic but also adds docs for Paddle Billing.

This also required adding a `payment_method_transaction` method on the `Paddle::Subscription` as this is used for when updating payment method details, as Paddle Billing doesn't use the `paddle_update_url` like Classic does.


deanpcmad:
I'd say this is ready :)

Issue #874, opened 2023-10-18
Paddle Billing transactions which include credit


deanpcmad:
With Paddle Billing, credits maybe added to a transaction such as if a subscription's plan is upgraded, a credit is added for pro-rated updates.

For example, here's a webhook event for a plan upgrade from $15 to $20

```json
{
  "event_id": "evt_abc123",
  "event_type": "transaction.completed",
  "occurred_at": "2023-10-18T10:56:33.696554Z",
  "notification_id": "ntf_abc123",
  "data": {
    "id": "txn_abc123",
    "items": [
      {
        "price": {
          "id": "pri_abc123",
          "status": "active",
          "quantity": {
            "maximum": 100,
            "minimum": 1
          },
          "tax_mode": "account_setting",
          "product_id": "pro_abc123",
          "unit_price": {
            "amount": "2000",
            "currency_code": "USD"
          },
          "custom_data": null,
          "description": "20 monthly",
          "trial_period": null,
          "billing_cycle": {
            "interval": "month",
            "frequency": 1
          },
          "unit_price_overrides": []
        },
        "price_id": "pri_abc123",
        "quantity": 1,
        "proration": {
          "rate": "0.99966",
          "billing_period": {
            "ends_at": "2023-11-18T10:41:13.871061Z",
            "starts_at": "2023-10-18T10:56:23.015Z"
          }
        }
      }
    ],
    "origin": "subscription_update",
    "status": "completed",
    "details": {
      "totals": {
        "fee": "150",
        "tax": "333",
        "total": "1999",
        "credit": "1500",
        "balance": "0",
        "discount": "0",
        "earnings": "1516",
        "subtotal": "1666",
        "grand_total": "499",
        "currency_code": "USD"
      },
      "line_items": [
        {
          "id": "txnitm_abc123",
          "totals": {
            "tax": "333",
            "total": "1999",
            "discount": "0",
            "subtotal": "1666"
          },
          "product": {
            "id": "pro_abc123",
            "name": "My Test Payments App",
            "status": "active",
            "image_url": null,
            "custom_data": null,
            "description": null,
            "tax_category": "standard"
          },
          "price_id": "pri_abc123",
          "quantity": 1,
          "tax_rate": "0.2",
          "proration": {
            "rate": "0.99966",
            "billing_period": {
              "ends_at": "2023-11-18T10:41:13.871061Z",
              "starts_at": "2023-10-18T10:56:23.015Z"
            }
          },
          "unit_totals": {
[...]

PR #875, opened 2023-10-18
Paddle Billing Updates

deanpcmad:
This PR updates some old `PaddlePay` code that was left during the transfer. 

It updates the `processor_subscription` and `charge` methods.

When creating a one-off charge on Paddle, it returns the subscription back and not the transaction, which is annoying. So I've left it so it creates the transaction and then the webhook will receive that as normal


deanpcmad:
Checking the Paddle docs, they recommend grabbing the latest transactions from the API when doing an immediate charge, so I'll add a commit for this shortly

excid3:
Thanks for this! 👍

deanpcmad:
Updated :+1:

lib/pay/paddle/billable.rb
```diff
@@ -56,7 +56,19 @@ def charge(amount, options = {})
           effective_from: "immediately"
         )
 
-        # This charge is then created from the webhook
+        transaction = ::Paddle::Transaction.list(subscription_id: subscription.processor_id).data.try(:first)
```
excid3, line 59: Creating a one-time charge required a subscription in Paddle Classic. I believe Paddle billing lets you create a one-time charge outside of a subscription now.

https://developer.paddle.com/api-reference/transactions/create-transaction

If so, we should change this to be consistent with the other integrations we have.

deanpcmad: Transactions can be created, but they don't seem to charge on creation even if `collection_mode` is set to `automatic` so I'm not sure if that's a bug on Paddle's end or if that is intended :thinking:

excid3: Their API is awfully confusing sometimes. 🙃

deanpcmad: Just noticed the reply I had from Kieran, head of dev experience at Paddle - https://twitter.com/k__mountford/status/1714729918326534235

Payment details are tied to a subscription, so this is the only way at the moment. They say payment methods at a customer level will be coming later, which would mean this could be changed

excid3: That's good news. In the meantime, maybe we should still switch this so it generates a Transaction and document that it needs to be passed to Checkout.

It seems far more common that people want to make one-time charges that don't require a subscription. Then we can improve this further when that feature rolls out.

deanpcmad:
Have updated the `charge` code so it now creates a Transaction and have updated the docs. Also pushed a fix for #874 as part of this too.

excid3:
@deanpcmad how much work do you think it would be to migrate PaddleClassic to your paddle gem?

deanpcmad:
Shouldn't be too much work as I think most of the methods are similar.  Will try and take a look tonight

excid3:
Figure it would simplify long-term maintenance, so that would be great. 👍 
Would like to roll it into our v2 release if it's easy enough.

PR #878, opened 2023-10-18
Add Rails 7.1 to tests and remove Ruby 2.7 as it's EOL

deanpcmad:


deanpcmad:
Some tests are failing as it appears it hit the rate-limit for Docker :laughing:

PR #886, opened 2023-10-26
Rename Paddle to Paddle Billing

deanpcmad:
I *think* I've renamed anything. I may tweak the Paddle docs and split them out into Billing and Classic, but will do that in another PR


deanpcmad:
I may tweak the Paddle docs and split them out into Billing and Classic

excid3:
Yeah, we'll want to do that. I also noticed we don't have any tests for paddle_billing so we will need to do those before we ship.

PR #890, opened 2023-10-29
Retrieve paddle subscription with options

samuelmolinari:
The paddle gem accept the option `extra` when retrieving a subscription.

See: https://github.com/deanpcmad/paddle/blob/83120a9e285cfb5daf36f7f56939e985ab029843/lib/paddle/models/subscription.rb#L11

What this PR allows us to do:

```ruby
subscription.processor_subscription(extra: 'next_transaction')
```

### DIFF
--- a/lib/pay/paddle_billing/billable.rb
+++ b/lib/pay/paddle_billing/billable.rb
@@ -81,9 +81,7 @@ module Pay
       end
 
       def processor_subscription(subscription_id, options = {})
-        opts = options.except(:id)
-
-        ::Paddle::Subscription.retrieve(id: subscription_id, **opts)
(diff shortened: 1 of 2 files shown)


samuelmolinari:
I've noticed this is not exactly what I wanted to achieve. `processor_subscription` can be called on a `Pay::Subscription` , and that's where we should be able to pass the wanted options:

```ruby
subscription.processor_subscription(extra: 'next_transaction')
```

Similar to how it's implemented on `Pay::Stripe::Subscription` (https://github.com/pay-rails/pay/blob/bd15f22f736e68571bda693bbe74103b90b35b45/lib/pay/stripe/subscription.rb#L159)

I'll work on that to complete this PR

excid3:
Maybe we should just include these extras by default?

PR #894, opened 2023-11-05
Remove PaddlePay

deanpcmad:
This PR removes PaddlePay for my Paddle library. Requested in a comment on #875


lib/pay/paddle_classic/subscription.rb
```diff
@@ -85,30 +84,30 @@ def cancel(**options)
         elsif paused?
           pause_starts_at
         else
-          Time.parse(processor_subscription.next_payment&.fetch(:date)) || Time.current
+          Time.parse(processor_subscription.next_payment&.try(:date)) || Time.current
```
excid3, line 87: In both cases here, I think this crashes on `Time.parse` if next_payment is nil. I think we pretty much always have a next_payment on an active subscription, so we can probably change this to require it.

```suggestion
          Time.parse(processor_subscription.next_payment.date)
```

I'm not sure about a `past_due` subscription though. That's probably the only situation we'd need to know for sure.

PR #895, opened 2023-11-05
Paddle Billing Client Tokens & docs tweaks

deanpcmad:
Paddle have recently added [client side tokens](https://developer.paddle.com/changelog/2023/client-side-tokens) as a way to interact with the Paddle API using Paddle.js without exposing your API tokens. Paddle recommend using these instead of passing the `seller` option, as new features may require it. 

The latest feature of Paddle.js is the ability to [generate localised price previews](https://developer.paddle.com/changelog/2023/paddle-js-pricing-pages) on the fly, which is great for building pricing pages as it shows the customer's local currency (with taxes too). This was a feature I've been requesting since the start as the previous version had it too.

This PR adds `client_token` to `PaddleBilling`, which is set by environment variables or in the credentials/secrets files. 

I've also updated the Paddle Billing subscription docs and renamed some `Pay::Paddle` examples to `Pay::PaddleClassic`

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -31,6 +31,8 @@ Pay::Customer.where(processor: :paddle).update_all(processor: :paddle_classic)
 You'll also need to update the Webhook endpoint from `/pay/webhooks/paddle` to `/pay/webhooks/paddle_classic`
 And rename custom webhooks from `paddle.*` to `paddle_classic.*`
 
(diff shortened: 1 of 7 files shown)


excid3:
I assume we should remove the seller_id too while we're here then since we haven't released this?

deanpcmad:
Ah nice, was just about to do that but I can see you've already done it 😅

PR #920, opened 2023-12-14
Safely handle Stripe payment methods without extra details

excid3:
I wasn't able to test this fix with BLIK payments on my account, but this should safely handle any payment methods that don't provide additional details.

@WozniakMac give this a try and see if it fixes your issue?


WozniakMac:
It works! 🎉 
Thank you!

PR #936, opened 2024-01-10
Clear trial end when subscription becomes active

excid3:
This fixes an issue with Paddle Classic where a subscription is synced after becoming `active` after `trialing` and the `trial_ends_at` time is not cleared.


excid3:
This probably also affects Paddle Billing.

PR #945, opened 2024-01-24
Update Paddle Billing Payment Methods to use `payment_method_id`

deanpcmad:
Paddle have added a new `payment_method_id` field, which returns a Paddle ID. This replaces the `stored_payment_method_id` field which returned a UUID.

For their changelog entry: https://developer.paddle.com/changelog/2024/payment-method-paddle-id

I assume this means they will be adding a Payment Methods API soon, which could be useful.

### DIFF
--- a/lib/pay/paddle_billing/payment_method.rb
+++ b/lib/pay/paddle_billing/payment_method.rb
@@ -19,7 +19,7 @@ module Pay
           attrs[:exp_year] = details.card.expiry_year
         end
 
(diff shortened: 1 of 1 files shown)


excid3:
This will probably produce duplicates I just realized since the UUID version of a payment method could be in the database already. I'm not sure we need to do anything in Pay but maybe a note in the UPGRADE guide for the next release?

deanpcmad:
Ah, good point. Yeah, adding that to the upgrade docs is a good idea.

PR #946, opened 2024-01-24
Fix to Paddle Billing Payment Method creation

deanpcmad:
This PR fixes a bug found in this discussion: https://github.com/pay-rails/pay/discussions/922#discussioncomment-8191211

It also adds a new method called `sync_from_transaction` which takes both a `pay_customer` and a `transaction`. This then uses the Paddle API to grab the details for that transaction and create/update a Payment Method for the given Pay Customer.

```ruby
Pay::PaddleBilling::PaymentMethod.sync_from_transaction pay_customer: Customer.last.payment_processor, transaction: "txn_01hmxtm1ep1c2pznfftybn56kk"
```

### DIFF
--- a/docs/4_payment_methods.md
+++ b/docs/4_payment_methods.md
@@ -72,6 +72,20 @@ You can also add a payment method without making it the default.
 @user.payment_processor.add_payment_method(params[:payment_method_token], default: false)
 ```
 
+## Importing Payment Methods
+
+### Paddle Billing
+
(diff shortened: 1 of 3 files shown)


j-m-harris:
Any movement on getting this merged? Also just hit this problem (thanks for having already prepared the fix though!)

mrmason:
+1 for having this merged if possible

excid3:
We should update tests for this, but I'm going to merge this anyways.

PR #898, opened 2023-11-07
Added support for Lemon Squeezy 

deanpcmad:
This PR adds support for Lemon Squeezy, a payment processor that acts as a Merchant of Record. You can see their site here > https://www.lemonsqueezy.com/

Lemon Squeezy are very similar to Paddle's new Billing APIs, which is one of the reasons I started work on adding it :sweat_smile: 

This code was originally written in July, but I've re-based it with the current changes. It still needs some work so I'm marking this as a draft. I also need to update the [Lemon Squeezy Gem](https://github.com/deanpcmad/lemonsqueezy) that I wrote too as I believe there's been some more API updates

See #897.

PR #947, opened 2024-01-24
Lemon Squeezy support

deanpcmad:
This PR adds support for Lemon Squeezy, a Merchant of Record payment provider. Their site is here > https://www.lemonsqueezy.com/

It uses a Ruby library I built, [lemonsqueezy](https://github.com/deanpcmad/lemonsqueezy).

This is a rewrite of #898 from what I've learnt with integrating Paddle Billing. The Lemon Squeezy API is quite similar to Paddle's Billing API. I have noticed a few things though:

- Trial Subscriptions have the status of `on_trial`. Not sure if you want to set it as that or replace the status with `trialing` when saving?
- They don't support cancelling subscriptions immediately.
- Card expiry dates aren't listed in the API, just the brand and last 4 
- Charges cannot be created from the API

This is a WIP.


harrison-broadbent:
Keen!

excid3:
`trialing` would be nice to make it consistent everywhere which is helpful for querying. Although I don't like modifying data from the API if we can help it. I would say we should map it.

We can raise an error for `cancel_now` if they don't support immediate cancellations.

What else do we need to implement?

deanpcmad:
Another oddity I've found, is that they don't list card expiry dates in the API, just the brand and last 4 🤔

excid3:
> Another oddity I've found, is that they don't list card expiry dates in the API, just the brand and last 4 🤔

That is very odd.

deanpcmad:
> `trialing` would be nice to make it consistent everywhere which is helpful for querying. Although I don't like modifying data from the API if we can help it. I would say we should map it.

So when a subscription is a trial, set the status as `trialing` or keep it as is?

deanpcmad:
Have just pushed a couple more commits, including initial work on docs. Feel free to give it a go.

I've noticed they don't have an API for creating charges, so I've sent them an email requesting this plus a few other minor details such as card expiry dates.

deanpcmad:
As for linking a subscription to a LS Customer, it appears that setting the email address in the URL links it, but I've added support for passthroughs just like Paddle Classic. This is another thing I've asked about, seeing if we can just pass the Customer ID to link it.

An example payment link I've been using has been in this format:

```
https://voupedev.lemonsqueezy.com/checkout/buy/xxx?embed=1&checkout[email]=yyy&checkout[custom][passthrough]=zzz
```

You should replace `xxx` with the UUID of the product, which can be viewed when clicking the Share button.

app/controllers/pay/webhooks/lemon_squeezy_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+      # Pass Lemon Squeezy signature from request.headers["X-Signature"]
+      def valid_signature?(signature)
+        return false if signature.blank?
+
+        key = Pay::LemonSqueezy.signing_secret
+        data = request.raw_post
+        digest = OpenSSL::Digest.new("sha256")
+
+        hmac = OpenSSL::HMAC.hexdigest(digest, key, data)
+        hmac == signature
```
kyrylo, line 37: You might want to use [ActiveSupport::SecurityUtils.secure_compare](https://api.rubyonrails.org/classes/ActiveSupport/SecurityUtils.html) here.

```suggestion
        ActiveSupport::SecurityUtils.secure_compare(hmac, signature)
```

deanpcmad:
Not sure why those tests are failing when they work for me 🤔

app/controllers/pay/webhooks/lemon_squeezy_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+        key = Pay::LemonSqueezy.signing_secret
+        data = request.raw_post
+        digest = OpenSSL::Digest.new("sha256")
+
+        hmac = OpenSSL::HMAC.hexdigest(digest, key, data)
+        ActiveSupport::SecurityUtils.secure_compare(hmac, signature)
+      end
+
+      def verify_params
+        params.except(:action, :controller).permit!
```
kyrylo, line 41: FYI, this line produces a Brakeman warning in my app (I don't use pay-rails; instead, I work with the lemonsqueezy library directly):

```
== Warnings ==

Confidence: Medium
Category: Mass Assignment
Check: MassAssignment
Message: Specify exact keys allowed for mass assignment instead of using `permit!` which allows any keys
Code: params.except(:action, :controller).permit!
Line: 33
```

deanpcmad: Yeah I don't think that should be an issue, as we verify the webhook signature with what is sent, before using that method.

excid3:
1. That's fine, just have the charge method raise an error?
2. I don't think we need the passthrough

excid3:
LemonSqueezy.js 2.0 is out now. https://github.com/lmsqueezy/lemonsqueezy.js/releases/tag/v2.0.0

deanpcmad:
Ha, I just received an email from them saying that 😅  I believe that's their official API library, not Lemon.js which is used for checkout overlays.

excid3:
Ohhh, like backend Node.js library, not client side? 🙃

deanpcmad:
Yeah 😆  - Lemon.js is the client-slide library - https://docs.lemonsqueezy.com/help/lemonjs/what-is-lemonjs

excid3:
And to think they could have called the Node SDK "lemonode.js". Such a shame.

deanpcmad:
Wow published 11 years ago 😅

nduplessis:
I would love to help get this over the line, is there anything still outstanding that's required for a merge?

joshfester:
Does this mean lemon squeezy support will be coming to Jumpstart? 👀

Gary-H9:
👋  Similar to @nduplessis, I'd like to help here if possible/necessary.

5andu:
Any updates or help needed for this? :)

deanpcmad:
I believe it's ready for people to test, so feel free to give it a go

5andu:
Thanks!

deanpcmad:
Ah good spot, will change that

lib/pay/lemon_squeezy/subscription.rb
```diff
@@ -0,0 +1,181 @@
...
+        :processor_subscription,
+        :prorate,
+        :prorate?,
+        :quantity,
+        :quantity?,
+        :trial_ends_at,
+        to: :pay_subscription
+
+      def self.sync_from_transaction(transaction_id)
+        transaction = ::Paddle::Transaction.retrieve(id: transaction_id)
```
k0va1, line 26: I suppose this method was copy-pasted from `paddle`, not quite sure that it should be here :)

k0va1:
> This means a passthrough isn't required, but it may be best to keep it for if a customer doesn't exist?

@deanpcmad I'm not quite sure if it's a valid approach but I'm looking for a customer like this https://github.com/k0va1/pay/pull/1/commits/4cae51006fd036a4ce87f76faf0c89b4e9139146

Passthrough might be not a bad option tho

deanpcmad:
Yeah, maybe it would be best to keep the passthrough there

lib/pay/lemon_squeezy/subscription.rb
deanpcmad: Yep it was, have removed this now

deanpcmad:
@k0va1 feel free to test it now. I've reverted the commit that removes the passthrough code & docs

deanpcmad:
@excid3 have you had chance to take a look at this yet? 😄

excid3:
A little, but need to build an end-to-end implementation to test it.

lib/pay/lemon_squeezy/charge.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      delegate :processor_id, :customer, to: :pay_charge
+
+      def initialize(pay_charge)
+        @pay_charge = pay_charge
+      end
+
+      def self.sync(charge_id, object: nil, try: 0, retries: 1)
+        # Skip loading the latest subscription invoice details from the API if we already have it
+        object ||= ::LemonSqueezy::SubscriptionInvoice.retrieve(id: charge_id)
```
k0va1, line 14: what if we don't find any invoice and got 404 error? i suppose this case should be handled

I bumped into an `Error 404: No results were found for your request. '[{"status"=>"404", "title"=>"Not Found"}]' (LemonSqueezy::Error)` while accidentally sending test webhook to production server

deanpcmad:
Well, Stripe just acquired Lemon Squeezy - https://www.lemonsqueezy.com/blog/stripe-acquires-lemon-squeezy

excid3:
Saw that, super exciting!

app/models/pay/webhook.rb
```diff
@@ -8,7 +8,7 @@ def process!
       Pay::Webhooks.instrument type: "#{processor}.#{event_type}", event: rehydrated_event
 
       # Remove after successfully processing
-      destroy
+      # destroy
```
nflorentin, line 11: I was wondering why the addition of lemon_squeezy is commenting this line of code.

PR #948, opened 2024-01-27
Address nillable Paddle Billing values in subscription

oliwoodsuk:
## Pull Request

**Summary:**
As listed [here](https://developer.paddle.com/api-reference/subscriptions/overview), Paddle subscription attributes `next_billed_at`, `paused_at` and  `scheduled_change` can all be `nil`. 
I ran into this issue when having Paddle set up in sandbox mode.

**Description:**
`Pay::Subscription.trial_ends_at` and `pause_starts_at` will fallback to `nil` is not present in the Paddle subscription.
`Pay::Subscription.ends_at` will default to the current date if not present, so the `canceled?` method works atleast.

### DIFF
--- a/lib/pay/paddle_billing/subscription.rb
+++ b/lib/pay/paddle_billing/subscription.rb
@@ -56,9 +56,9 @@ module Pay
           # Remove payment methods since customer cannot be reused after cancelling
           Pay::PaymentMethod.where(customer_id: object.customer_id).destroy_all
         when "trialing"
-          attributes[:trial_ends_at] = Time.parse(object.next_billed_at)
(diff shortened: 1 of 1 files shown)


lib/pay/paddle_billing/subscription.rb
```diff
@@ -109,7 +109,7 @@ def cancel(**options)
         )
         pay_subscription.update(
           status: response.status,
-          ends_at: response.scheduled_change.effective_at
+          ends_at: response.scheduled_change&.effective_at || DateTime.now
```
excid3, line 112:
```suggestion
          ends_at: response.scheduled_change&.effective_at || Time.current
```

Issue #991, opened 2024-05-29
[BUG] Unpausing a Stripe subscription


rares-c:
## Bug Report

Users cannot unpause a Stripe subscription. The Stripe subscription is marked as `unpaused` locally, but in Stripe the subscription remains unpaused. The parameter `pause_collection: nil` is not passed to Stripe in the POST request when trying to unpause a subscription.

**To Reproduce:**
1. Pause a Stripe subscription
2. Resume the subscription
3. Check Stripe to see that the subscription is still paused
4. Check your app to see that the subscription is no longer paused.

**Expected Behavior:**
The subscription would unpause in Stripe as well.

**Actual Behavior:**
The subscription remains paused in Stripe and unpaused locally.

**Environment:**
- Pay gem version: 6.8.1
- Ruby version: 3.2.2
- Rails version: 7.1.3.3
- Operating System: MacOS

**Possible Fix:**
Change `pause_collection: ""` instead of `pause_collection: nil` in the `unpause` method of the `Pay::Stripe::Subscription` class.

**Labels to Apply:**
Bug

PR #992, opened 2024-05-29
Fixed unpause stripe subscription empty params bug

rares-c:
## Pull Request

**Summary:**
Fixes the bug where users cannot unpause a Stripe subscription.

**Related Issue:**
#991

**Description:**
Changed the `pause_collection: ""` instead of `pause_collection: nil` in the `unpause` method of the `Pay::Stripe::Subscription` class so that it can be passed on to Stripe.


excid3:
We will need a test for this.

PR #1004, opened 2024-06-11
Fix: Unset previous default payment methods when adding a new one using FakeProcessor

f19ps:
## Pull Request

**Summary:**
When using `FakeProcessor`, adding a new `default` payment method does not unset the previous default payment methods, causing the `payment_processor.default_payment_method` method to return an outdated default payment method.


excid3:
Looks good to me. 👍

PR #1009, opened 2024-06-19
Handle Paddle billing customer already exists error

nduplessis:
## Pull Request

**Summary:**
When attempting to add a customer to Paddle Billing that already exists, an unhandled error occurs

**Description:**
https://developer.paddle.com/errors/customers/customer_already_exists

**Testing:**

**Checklist:**

- [ ] Code follows the project's coding standards
- [ ] All existing tests pass

**Additional Notes:**


excid3:
Need to have a test for this.

We should also probably update @deanpcmad's paddle gem to raise different types of errors so we can rescue from this specifically.

nduplessis:
Yeah sorry about that, first time ever saving a PR as a draft and I didn't really expect it to show up here tbh. Will be adding a test for this soon and see if I can get a change to the paddle gem as suggested

excid3:
No worries! You're good. Was mostly just making some notes. 👍

lib/pay/paddle_billing/billable.rb
```diff
@@ -33,7 +33,13 @@ def customer
           sc
         end
       rescue ::Paddle::Error => e
-        raise Pay::PaddleBilling::Error, e
+        sc = Paddle::Customer.list(email:).data.first
```
excid3, line 36: Paddle's CustomerAlreadyExists error message contains the ID in it, so we can parse it out of the string I think:

> "customer email conflicts with customer of id %s"

excid3: Can always do that as a fallback if it can't find the value. 👍

deanpcmad:
> We should also probably update @deanpcmad's paddle gem to raise different types of errors so we can rescue from this specifically.

Good idea 👍 
There are [quite a few](https://developer.paddle.com/errors/overview), so I won't want to manually add them all, but feel free to submit a PR for that

shettytejas:
@nduplessis Is this PR still in development? This would be a great addition!

nduplessis:
Hi yes it is, just had some deadlines I needed to get to. Will have this wrapped up before the end of the weekend!

shettytejas:
@nduplessis That's great news! Thanks for the prompt reply!

nduplessis:
@deanpcmad I've created a PR for this at https://github.com/deanpcmad/paddle/pull/6, let me know what you think of that approach please

excid3:
This seems to have introduced a regression by changing the `def initialize(msg = nil)` signature for `Paddle::Error` as it now requires two arguments. Also Paddle Classic didn't get updated to match?

lib/pay/paddle_billing/billable.rb
```diff
@@ -33,7 +33,13 @@ def customer
...
+        if sc.present?
+          pay_customer.update!(processor_id: sc.id)
```
excid3, line 38: I'm not quite sure this logic makes sense. We would end up with Pay::Customer records with duplicate `processor_id` here, so we still need to raise an error here.

deanpcmad:
With Paddle v2.5, it will now return a `Paddle::Errors::ConflictError` error when a customer already exists, so that can be integrated into this PR

excid3:
Yeah, I'm just not sure what Pay should do in this case. Probably just continue raising the error as it currently does I guess?

deanpcmad:
Yeah probably? Then a doc update to say it could return that error?

excid3:
Yeah, probably want to just close this and mark it as a limitation in the Paddle overview docs. Same thing applies to Lemon Squeezy I'm pretty sure.

PR #1027, opened 2024-07-29
Lock Pay::Customer while creating a Stripe customer

mpvosseller:
## Pull Request

**Summary:**

Lock the `Pay::Customer` while creating its associated Stripe customer. This is to prevent a race condition where more than one Stripe customers can be created for a single `Pay::Customer`.

**Related Issue:**

See discussion: https://github.com/pay-rails/pay/discussions/857

**Description:**

We have twice now encountered a race condition where more than one Stripe customer was created for a single `Pay::Customer`. This was discovered because of bugs in our React frontend which caused it to accidentally fire multiple API requests to the backend at the same time. Two requests (two threads) would enter the `customer` method at the same time. Both would see that no Stripe customer yet existed and both would begin creating one. The last one to complete would overwrite the previous one in the database. This was especially bad in our app because we use Stripe checkout. In our case our user ended up checking out with a Stripe customer that was no longer associated with their `Pay::Customer`. With this change we now lock the `Pay::Customer` while creating a Stripe customer to prevent more than one from being created.

The change simply updates the `customer` method by moving the existing code inside a top level `with_lock` block on the `Pay::Customer`.

**Testing:**

I didn't add any new tests as I didn't see an existing pattern for verifying this type of locking code (which is difficult) or an existing tests for similar locking code. I did some manual testing though (using log statements and sleep statements) to verify that only one thread will run the critical section at a time.

**Checklist:**

- [X] Documentation has been updated (if applicable)
- [X] All existing tests pass

**Additional Notes:**

### DIFF
--- a/lib/pay/stripe/billable.rb
+++ b/lib/pay/stripe/billable.rb
@@ -48,22 +48,24 @@ module Pay
       #
       # Returns a Stripe::Customer object
       def customer
-        stripe_customer = if processor_id?
-          ::Stripe::Customer.retrieve({id: processor_id, expand: ["tax", "invoice_credit_balance"]}, stripe_options)
-        else
-          sc = ::Stripe::Customer.create(customer_attributes.merge(expand: ["tax"]), stripe_options)
-          pay_customer.update!(processor_id: sc.id, stripe_account: stripe_account)
-          sc
-        end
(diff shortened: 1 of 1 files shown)


excid3:
This would only protect in the case that you already have a `Pay::Customer` record in the database, not the situation where it's lazy created, right?

PR #1042, opened 2024-08-09
Use STI for Pay models

excid3:
The current organization of payment processor logic is separated out from the models which has felt clunky.

By using STI, we can automatically load the correct payment processor classes and simplify the codebase significantly.


deanpcmad:
Sounds like a good idea 👀

excid3:
Saved a couple hundred lines of code, not too shabby.

PR #1045, opened 2024-08-13
Update Stripe.js documentation link

Mth0158:
Looks like Stripe change the previous link is not the right one (it's redirecting to [https://docs.stripe.com/payments/elements](https://docs.stripe.com/payments/elements))


excid3:
Thanks @Mth0158 👍

PR #1051, opened 2024-08-28
Add support for Stripe's Meter Events for usage based billing

ayushn21:
## Pull Request

**Summary:**
Usage records for usage based billing is now a legacy method in Stripe. Meter events are the new way to track usage: https://docs.stripe.com/api/billing/meter-event?lang=ruby.

**Description:**
I've added a method to make an API call to Stripe to create a meter event. The response from the Stripe SDK is returned to the caller to handle as needed.

**Checklist:**

- [x] All existing tests pass

### DIFF
--- a/app/models/pay/stripe/subscription.rb
+++ b/app/models/pay/stripe/subscription.rb
@@ -301,6 +301,21 @@ module Pay
         raise Pay::Stripe::Error, e
       end
 
+      # Creates a meter event to bill for usage
+      #
+      # create_meter_event(:api_request, value: 1)
+      # create_meter_event(:api_request, token: 7)
+      def create_meter_event(event_name, **payload)
+        api_record unless processor_id?
+        ::Stripe::Billing::MeterEvent.create({
+          event_name: event_name,
+          payload: {
+            stripe_customer_id: processor_id,
+            **payload
+          }
+        })
+      end
+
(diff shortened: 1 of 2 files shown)


app/models/pay/stripe/subscription.rb
```diff
@@ -301,6 +301,20 @@ def swap(plan, **options)
...
+      # Creates a meter event to bill for usage
+      #
+      # create_meter_event(:api_request, value: 1)
+      # create_meter_event(:api_request, token: 7)
+      def create_meter_event(event_name, **payload)
+        ::Stripe::Billing::MeterEvent.create({
+          event_name: event_name,
+          payload: {
+            stripe_customer_id: customer.processor_id,
+            **payload
```
excid3, line 313: This would hit the API twice (one to load the Stripe::Customer, one to create the MeterEvent), so we should only do that if the processor_id does not exist since we lazy create customers.

```suggestion
        api_record unless processor_id?
        ::Stripe::Billing::MeterEvent.create({
          event_name: event_name,
          payload: {
            stripe_customer_id: processor_id,
            **payload
```

ayushn21: Ahh thanks, I missed that `processor_id` was available on the subscription model itself. Updated now.

PR #1081, opened 2024-09-30
Explicitly require ostruct gem in webhook model

jjatinggoyal:
## Pull Request

**Summary:**
Running into `Error: NameError: uninitialized constant Pay::Webhook::OpenStruct` error when `Pay::Webhooks::ProcessJob` is running for `paddle_billing`

**Description:**
I'm integrating paddle with pay gem into my application, and have configured multiple webhooks for syncing subscriptions.

However, I'm running into the following error in `Pay::Webhooks::ProcessJob`:

```
Error: NameError: uninitialized constant Pay::Webhook::OpenStruct
/usr/local/bundle/ruby/3.3.0/gems/pay-8.0.0/app/models/pay/webhook.rb:35:in `to_recursive_ostruct'
/usr/local/bundle/ruby/3.3.0/gems/pay-8.0.0/app/models/pay/webhook.rb:21:in `rehydrated_event'
/usr/local/bundle/ruby/3.3.0/gems/pay-8.0.0/app/models/pay/webhook.rb:8:in `process!'
/usr/local/bundle/ruby/3.3.0/gems/pay-8.0.0/lib/pay/webhooks/process_job.rb:5:in `perform'
```

To fix the issue, I have explicitly added `require "ostruct"` to the `Pay::Webhook` model, and added a test case for `paddle_billing` for the same model.

Additionally, added the `require "ostruct"` line to `Pay::Braintree::Subscription` model (the only other place where OpenStruct is used) as well proactively.

**Testing:**
Tested the fixes in this PR on my staging server, the webhook job passes now.

Screenshot 2024-10-01 at 12 50 26 AM

The first (failed) run in above screenshot is with pay-8.0.0, while the second successful run is with the fixes in this PR.

**Additional Notes:**
My setup is:
- Rails 7.2.1
- Ruby 3.3.4
- Pay 8.0.0
- Paddle 2.6.0

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -11,6 +11,7 @@ PATH
   remote: .
   specs:
     pay (8.0.0)
+      ostruct
       rails (>= 6.0.0)
 
 GEM
--- a/app/models/pay/braintree/subscription.rb
+++ b/app/models/pay/braintree/subscription.rb
@@ -1,3 +1,5 @@
+require "ostruct"
+
 module Pay
   module Braintree
     class Subscription < Pay::Subscription
--- a/app/models/pay/webhook.rb
+++ b/app/models/pay/webhook.rb
@@ -1,3 +1,5 @@
+require "ostruct"
+
 module Pay
   class Webhook < Pay::ApplicationRecord
     validates :processor, presence: true
--- a/gemfiles/rails_6_1.gemfile
+++ b/gemfiles/rails_6_1.gemfile
@@ -26,5 +26,6 @@ gem "net-imap", require: false
 gem "net-pop", require: false
 gem "net-smtp", require: false
 gem "rails", "~> 6.1.0"
+gem "ostruct"
 
 gemspec path: "../"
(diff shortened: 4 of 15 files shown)


excid3:
We may want to just get rid of our OpenStruct since we don't use it much.

It's moving out of Ruby core to a bundled gem in 3.5: https://bugs.ruby-lang.org/issues/20309

Could possibly replace it with `Data` or something Rails provides.

jjatinggoyal:
Just read up on `Data` class, look like it requires you to declare the class attributes beforehand, which doesn't fit our use case here one-to-one.

As for rails, maybe `ActiveSupport::HashWithIndifferentAccess` could be used, but the access pattern is different than ostruct. Do you have anything specific in mind?

excid3:
Might not be worth messing with yet.

We should also add `ostruct` as a dependency in the gemspec as it will be removed from Ruby in 3.5.

jjatinggoyal:
Cool, let me add `ostruct` as a dependency. Won't mention a min version, hope that is fine.

jjatinggoyal:
@excid3 does this PR look good to you? Can we merge it?

excid3:
I figured out a way to remove OpenStruct and use a feature built-in to Rails instead.

PR #1120, opened 2024-12-20
Adding self.sync to fake processor for a consistent API

luizkowalski:
## Pull Request

**Summary:**
Adding `self.sync` to fake processor for API consistency

**Description:**
Currently, the fake processor does not implement `self.sync`, so calling `sync!` if a user is subscribed to a `fake` process (in my case, users whom I gave free access) triggers an error. Ideally, since it does nothing, it should just be a bypass so that clients don't need to check the subscription type before calling the `sync!` or not

Also fixed a couple of typos while going through the upgrade guide. 

**Testing:**

**Checklist:**

- [X] Tests have been added or updated to cover the changes
- [ ] All existing tests pass

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -110,7 +110,7 @@ This is a major change to add Stripe tax support, Stripe metered billing, new co
 
 ### Method Additions and Changes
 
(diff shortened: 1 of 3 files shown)


excid3:
Appreciate this @luizkowalski!

PR #1129, opened 2024-12-30
Update stripe_connect.md to reflect Stripe API changes and offer some color on Merchant#onboarding_complete

TrevorHinesley:
Updated docs to provide some detail on the default `onboarding_complete` flag that's set based on account verification, and also felt it worth a callout about the gotcha related to Stripe Connect's API updates. Also updated syntax that might have been outdated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Unreleased
 
+* Remove default `type` for Stripe Merchant Account creation
+
 ### 8.3.0
 
 * Ignore Stripe `payment_failed` and `payment_action_required` webhooks on `incomplete` subscriptions as these are already handled by the JavaScript in-browser. #1121
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Unreleased
 
-* Remove default `type` for Stripe Merchant Account creation
+* [Breaking] Remove default `type` for Stripe Merchant Account creation
 
 ### 8.3.0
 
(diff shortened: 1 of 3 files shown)


docs/marketplaces/stripe_connect.md
```diff
@@ -33,8 +33,11 @@ end
 @user.merchant_processor.account_link
 @user.merchant_processor.login_link
 @user.merchant_processor.transfer(amount: 25_00)
+@user.merchant_processor.onboarding_complete? # Updates via webhook based on the Stripe::Account's #charges_enabled attribute
 ```
 
+**IMPORTANT:** [Stripe has updated their account creation mechanism](https://docs.stripe.com/connect/migrate-to-controller-properties) to offer more granularity by providing a `controller` parameter. If you want to use specific `controller` options when creating an account, you'll instead need to map your desired settings to a `type` [following this Stripe guide](https://docs.stripe.com/connect/migrate-to-controller-properties#mapping-account-types-to-controller-parameters). At the moment, Pay defaults to specifying a `type` in the `Merchant#create_account` method, and both parameters cannot be sent simultaneously to Stripe. If you need to use `controller` to have more granular settings, we welcome a PR.
```
excid3, line 39: I think it makes sense to just remove the default `type` since we can't really make any assumptions about how you plan to use it.

TrevorHinesley: Good with me, just wasn’t sure if that was important to remain or would be a breaking change for existing users

excid3: It'll be a breaking change, but I'm alright with that.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### Unreleased
 
+* Remove default `type` for Stripe Merchant Account creation
```
excid3, line 5:
```suggestion
* [Breaking] Remove default `type` for Stripe Merchant Account creation
```

PR #1142, opened 2025-03-14
Fix documentation for fetching API record from processor

kengreeff:
## Pull Request

**Summary:**
Just a small change to fix docs

**Description:**
Current documentation doesn't work - undefined method. Added the correct method.


excid3:
Good catch!

PR #1157, opened 2025-05-09
Lock set_payment_processor to prevent accidentally creating multiple pay_customers for a user

mpvosseller:
## Pull Request

**Summary:**

This adds use of a lock in the `set_payment_processor` to prevent accidentally creating more than one pay_customer for a user if it is called more than once at the same time (from multiple threads).

**Related Issue:**

https://github.com/pay-rails/pay/issues/1158

**Description:**

**Testing:**

**Checklist:**

- [ ] Code follows the project's coding standards
- [ ] All existing tests pass

**Additional Notes:**

### DIFF
--- a/lib/pay/attributes.rb
+++ b/lib/pay/attributes.rb
@@ -81,7 +81,7 @@ module Pay
       end
 
       def set_merchant_processor(processor_name, **attributes)
-        ActiveRecord::Base.transaction do
+        with_lock do
           pay_merchants.update_all(default: false)
           pay_merchant = pay_merchants.where(processor: processor_name, type: "Pay::#{processor_name.to_s.classify}::Merchant").first_or_initialize
           pay_merchant.update!(attributes.merge(default: true))
--- a/lib/pay/attributes.rb
+++ b/lib/pay/attributes.rb
@@ -29,22 +29,20 @@ module Pay
       # - Removes the default flag from all other Pay::Customers
       # - Removes the default flag from all Pay::PaymentMethods
       def set_payment_processor(processor_name, allow_fake: false, **attributes)
+        raise Pay::Error, "Processor `#{processor_name}` is not allowed" if processor_name.to_s == "fake_processor" && !allow_fake
+
+        # Safety check to make sure this is a valid Pay processor
+        klass = "Pay::#{processor_name.to_s.classify}::Customer".constantize
+        raise ArgumentError, "not a valid payment processor" if klass.ancestors.exclude?(Pay::Customer)
+
         with_lock do
-          raise Pay::Error, "Processor `#{processor_name}` is not allowed" if processor_name.to_s == "fake_processor" && !allow_fake
-
-          # Safety check to make sure this is a valid Pay processor
-          klass = "Pay::#{processor_name.to_s.classify}::Customer".constantize
-          raise ArgumentError, "not a valid payment processor" if klass.ancestors.exclude?(Pay::Customer)
-
-          ActiveRecord::Base.transaction do
-            pay_customers.update_all(default: false)
-            pay_customer = pay_customers.active.where(processor: processor_name, type: klass.name).first_or_initialize
-            pay_customer.update!(attributes.merge(default: true))
-          end
-
-          # Return new payment processor
-          reload_payment_processor
+          pay_customers.update_all(default: false)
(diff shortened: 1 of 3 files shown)


lib/pay/attributes.rb
```diff
@@ -29,20 +29,22 @@ module CustomerExtension
...
+          # Safety check to make sure this is a valid Pay processor
+          klass = "Pay::#{processor_name.to_s.classify}::Customer".constantize
+          raise ArgumentError, "not a valid payment processor" if klass.ancestors.exclude?(Pay::Customer)
 
-        ActiveRecord::Base.transaction do
-          pay_customers.update_all(default: false)
-          pay_customer = pay_customers.active.where(processor: processor_name, type: klass.name).first_or_initialize
-          pay_customer.update!(attributes.merge(default: true))
-        end
+          ActiveRecord::Base.transaction do
```
excid3, line 39: `with_lock` starts a transaction, so this should no longer be necessary.

https://api.rubyonrails.org/classes/ActiveRecord/Locking/Pessimistic.html#method-i-with_lock

mpvosseller: @excid3 ah, great. Just updated it to remove the extra transaction. Now the PR simply replaces `ActiveRecord::Base.transaction` with `with_lock`. Tested locally and works for me.

excid3: Awesome. I think this did break a test though, not sure why.

mpvosseller: @excid3 I looked into this a little. It is failing in [`assert_indexed_selects` ](https://github.com/pay-rails/pay/blob/ec4e870d91161b6e752eaee87eb0233dd76a5fea/test/test_helper.rb#L86) which looks like it is trying to ensure an index is used in each select statement by inspecting the query plan (`explain`) and ensuring it contains the word `index`. 

The first failure is in SQLite which performs a new extra select but it's plan doesn't include the word `index`:
```
5|0|33|search users using integer primary key (rowid=?)
```

Not sure if this test should be reliable.  Looks like `assert_indexed_selects` is used twice in this app (`set merchant processor` and `set payment processor`. 

Is it worth trying to fix or should we just remove it?

excid3: `with_lock` reloads the record with a lock before running the transaction, so it's catching the reload SELECT that is added now instead of the original select. It might be doable to refactor so the assertion aggregates the results and asserts after they've all run rather than each select.

`set_merchant_processor` probably also needs to be updated.

mpvosseller: @excid3 Sorry, I'm not sure I understand the aggregation idea.

As far as I can tell the core issue is that `assert_indexed_selects` assumes the explain string of every SELECT query using an index will include the substring `index` and I'm not sure that is correct.   

In SQLite:

The extra sql query is:
```
SELECT "users".* FROM "users" WHERE "users"."id" = ? LIMIT ? 
```
It's plan / explain string is:
```
5|0|0|search users using integer primary key (rowid=?)
```
I would assume that does use an index but the assertion fails because the plan string does NOT have the substring `index`.

In MySQL

The extra sql query is:
```
SELECT `users`.* FROM `users` WHERE `users`.`id` = 1066402000 LIMIT 1 FOR UPDATE
```

It's plan / explain string is:
```
+----+-------------+-------+------------+-------+---------------+---------+---------+-------+------+----------+-------+
| id | select_type | table | partitions | type  | possible_keys | key     | key_len | ref   | rows | filtered | extra |
+----+-------------+-------+------------+-------+---------------+---------+---------+-------+------+----------+-------+
|  1 | simple      | users | null       | const | primary       | primary | 8       | const |    1 |    100.0 | null  |
+----+-------------+-------+------------+-------+---------------+---------+---------+-------+------+----------+-------+
1 row in set (0.00 sec)
```

The assertion fails again because this plan string does not have the substring `index` either.

One thing that makes the tests pass is to check that the explain string contains EITHER `index` OR `primary` but I really have no idea if that is a good / correct / reliable implementation either as I really don't know much about this stuff.

Thoughts? Can you elaborate on the aggregation idea?

mpvosseller: Thank you @excid3. Sorry I couldn't be more helpful on that.

excid3: You're good! I know we added it at one point because something was inefficient, but this code doesn't change hardly ever anymore so I think we're fine.

Issue #1166, opened 2025-05-24
Pay Associations Name Consistency?


donnfelker:
I'm running into some issues with name conflicts that are stemming from Pay, and I'm wondering if I'm going to have to just pull Pay completely out until I can figure out a work around as it impacts the system design I have. However, this would have been averted had some associations had different names. 

In this file: https://github.com/pay-rails/pay/blob/main/lib/pay/attributes.rb

We have these associations: 

**EXISTING**
```ruby
has_many :pay_customers, class_name: "Pay::Customer", as: :owner, inverse_of: :owner
has_many :charges, through: :pay_customers, class_name: "Pay::Charge"
has_many :subscriptions, through: :pay_customers, class_name: "Pay::Subscription"
has_one :payment_processor, -> { where(default: true, deleted_at: nil) }, class_name: "Pay::Customer", as: :owner, inverse_of: :owner
```

It would be great if the middle two (`:charges` and `:subscriptions`) and perhaps the last one `:payment_processor`) should also be prefixed with `pay_` for the association name like the first one does, like this:

**NEW**
```ruby
has_many :pay_customers, class_name: "Pay::Customer", as: :owner, inverse_of: :owner
has_many :pay_charges, through: :pay_customers, class_name: "Pay::Charge"
has_many :pay_subscriptions, through: :pay_customers, class_name: "Pay::Subscription"
has_one :pay_payment_processor, -> { where(default: true, deleted_at: nil) }, class_name: "Pay::Customer", as: :owner, inverse_of: :owner
```

The reason is, my application has its own concept of "subscriptions" and they need to be scoped to the `account` model (as this is in Jumpstart Pro). 

I realize this is a fundamental breaking change to the gem if it were to get introduced, but I could see others having a similar problem as a "subscription" is very overloaded. 

e.g. - Could be an email payment subscription (as it is here), an email list subscription, a "subscribe" like button you see on YouTube type of "subscription", a replenishment subscription, etc.

excid3:
Closed by #1173

PR #1167, opened 2025-05-24
Add prefix to associations and methods in `attributes.rb`

donnfelker:
**Disclaimer:** I understand this is a massive breaking change, so I empathize if there is some resistance to it.

## Pull Request

**Summary:**
Renames the associations and methods that are in `attributes.rb` to include the `pay_` prefix for consistency and to avoid common association name conflicts. 

Core changes in `attributes.rb`: 
- `has_many :charges` → `has_many :pay_charges`
- `has_many :subscriptions` → `has_many :pay_subscriptions`
- `has_one :payment_processor` → `has_one :pay_payment_processor`
- `set_payment_processor` → `set_pay_payment_processor`
- `add_payment_processor` → `add_pay_payment_processor`

**Related Issue:**
https://github.com/pay-rails/pay/issues/1166

**Description:**
From issue #1166
I'm running into some issues with name conflicts that are stemming from Pay, and I'm wondering if I'm going to have to just pull Pay completely out until I can figure out a work around as it impacts the system design I have. However, this would have been averted had some associations had different names. 

In this file: https://github.com/pay-rails/pay/blob/main/lib/pay/attributes.rb

We have these associations: 

**EXISTING**
```ruby
has_many :pay_customers, class_name: "Pay::Customer", as: :owner, inverse_of: :owner
has_many :charges, through: :pay_customers, class_name: "Pay::Charge"
has_many :subscriptions, through: :pay_customers, class_name: "Pay::Subscription"
has_one :payment_processor, -> { where(default: true, deleted_at: nil) }, class_name: "Pay::Customer", as: :owner, inverse_of: :owner
```

It would be great if all of the associations used a `pay_` prefix for the association name like the first one does, like this:

**NEW**
```ruby
has_many :pay_customers, class_name: "Pay::Customer", as: :owner, inverse_of: :owner
has_many :pay_charges, through: :pay_customers, class_name: "Pay::Charge"
has_many :pay_subscriptions, through: :pay_customers, class_name: "Pay::Subscription"
has_one :pay_payment_processor, -> { where(default: true, deleted_at: nil) }, class_name: "Pay::Customer", as: :owner, inverse_of: :owner
```

The reason is, my application has its own concept of "subscriptions" and they need to be scoped to the `account` model (as is done in Jumpstart Pro). 

I realize this is a fundamental breaking change to the gem if it were to get introduced, but I could see others having a similar problem as a "subscription" is very overloaded. 

e.g. - An app's use case might vary for what it considers a "subscription".  This could be a payment subscription (as it is here), an email list subscription, a "subscribe" like button you see on YouTube type of "subscription", a replenishment subscription, etc. 

**Testing:**
Updated all tests. 

n/a

**Checklist:**
- [x] All existing tests pass

**Additional Notes:**
n/a


excid3:
I've been wanting to rename the `charges` and `subscriptions` associations, but I don't know that we need to change the `payment_processor` ones. It's unlikely those will conflict with anything.

excid3:
I think `payment_processor` is much less likely to conflict and you could really skip `pay_customer` entirely if needed.

PR #1170, opened 2025-06-02
Don't modify trial_ends_at in the cancel_now method unless the subscription is actually on trial

mpvosseller:
## Pull Request

**Summary:**

This updates `Pay::Stripe::Subscription#cancel_now!` so it will not set (modify) the `trial_ends_at` property unless the subscription is currently `on_trial?`

We had been using `trial_ends_at` to indicate whether a subscription was a trial but found this was not working on subscriptions where  `cancel_now!` was called.

**Related Commit**
https://github.com/pay-rails/pay/commit/5a719ad90cf4afb2d28419233ab7575df5daec75

**Related Issue:**

**Description:**

**Testing:**

I'm not quite sure if / how I can add a new VCR cassette? Is that something you would do or can you provide me a reference / pointer to do so?

**Checklist:**

- [X] Documentation has been updated (if applicable)
- [X] All existing tests pass

**Additional Notes:**

### DIFF
--- a/app/models/pay/stripe/subscription.rb
+++ b/app/models/pay/stripe/subscription.rb
@@ -178,7 +178,7 @@ module Pay
         return if canceled? && ends_at.past?
 
         @api_record = ::Stripe::Subscription.cancel(processor_id, options.merge(expand_options), stripe_options)
-        update(trial_ends_at: Time.current, ends_at: Time.current, status: :canceled)
+        update(trial_ends_at: on_trial? ? Time.current : trial_ends_at, ends_at: Time.current, status: :canceled)
       rescue ::Stripe::StripeError => e
         raise Pay::Stripe::Error, e
       end
--- a/test/pay/stripe/subscription_test.rb
+++ b/test/pay/stripe/subscription_test.rb
@@ -54,10 +54,27 @@ class Pay::Stripe::SubscriptionTest < ActiveSupport::TestCase
       subscription = @pay_customer.subscribe(name: "default", plan: "default", trial_period_days: 14)
       assert subscription.active?
       assert subscription.on_trial?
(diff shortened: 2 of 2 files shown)


excid3:
Hmm, why are you cancelling the subscription immediately instead of on period end?

mpvosseller:
> Hmm, why are you cancelling the subscription immediately instead of on period end?

I guess the most common general use case would be when a customer "forgets to cancel" and customer support graciously decides to "cancel_now" and issue a partial refund.

Our company also had an unusual use case where we called `cancel_now!` on a bunch of subscriptions to migrate them to a totally different payment model (outside of the pay gem and stripe). (e.g. a b2b model where a business like the customer's employer pays for the access not the customer).

I assumed setting `trial_ends_at` in `cancel_now!` was just an oversight. If you don't think we should be relying on `trial_ends_at` as an indication of the subscription having been a trial we could probably use the Stripe object's  `trial_end` property instead now that the full Stripe object is in the database.

mpvosseller:
Sorry if I'm not being clear. Let me rephrase the problem.

Let's say a user creates a regular (non-trial) subscription and at some point we call `subscription.cancel` (to cancel at end of billing period) the `trial_ends_at` property stays nil forever.

Now let's say another user creates a regular (non-trial) subscription and at some point we call `subscription.cancel_now` (to cancel immediately) the `trial_ends_at` gets set to the current time.

Why does the pay gem set `trial_ends_at` when canceling with one method (cancel_now) but not with the  other (cancel)? And when neither subscription was a trial?

excid3:
The current implementation is also incorrect. The values set should match what's returned from the API since the webhooks will update the values anyways. Currently this will set different times for each.

PR #1186, opened 2025-07-31
Add owner association to charge, subscription, and payment method models

excid3:
This makes handling things like fulfillment after a charge or subscription is created easier so the user doesn't have to jump through the customer association.

```ruby
module ChargeExtensions
  extend ActiveSupport::Concern

  included do
    after_create_commit :create_license
  end

  def create_license
    # previously: `customer.owner.create_license`
    owner.create_license(course_id: metadata.dig("course_id"))
  end
end

Rails.configuration.to_prepare do
  Pay::Charge.include ChargeExtensions
end
```

### DIFF
--- a/app/models/pay/charge.rb
+++ b/app/models/pay/charge.rb
@@ -3,6 +3,7 @@ module Pay
     # Associations
     belongs_to :customer
     belongs_to :subscription, optional: true
+    has_one :owner, through: :customer
 
     # Scopes
     scope :sorted, -> { order(created_at: :desc) }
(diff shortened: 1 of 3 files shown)


mhenrixon:
Hmm, but now I get:

```
Warning: Ignoring invalid association :owner on Pay::Subscription (Cannot have a has_many :through association 'Pay::Subscription#owner' on the polymorphic object 'Owner#owner' without 'source_type'. Try adding 'source_type: "Owner"' to 'has_many :through' definition.)
```

PR #1191, opened 2025-08-24
Fix mail_arguments and mail_to configuration ArgumentError

marckohlbrugge:
## Summary

This PR fixes a critical bug where `mail_arguments` and `mail_to` configurations fail with `ArgumentError` when using the documented syntax.

## The Problem

The Pay gem switched to using `instance_exec` for `mail_arguments` and `mail_to` lambdas to provide better access to mailer instance methods. However, there was a discrepancy between the implementation and documentation:

- **Implementation**: Uses `instance_exec` which passes 0 arguments
- **Documentation**: Shows lambdas expecting 2 arguments `(mailer, params)`
- **Tests**: Were testing direct lambda calls instead of the actual runtime behavior

This caused runtime errors when following the official documentation:
```
config/initializers/pay.rb:34:in `block (2 levels) in <main>': wrong number of arguments (given 0, expected 2) (ArgumentError)
```

## The Solution

This PR adds backward compatibility by checking the lambda's arity:
- **Arity 0**: Use `instance_exec` (new preferred syntax)
- **Arity 2**: Call with explicit arguments (backward compatibility)

Both syntaxes now work:

```ruby
# New syntax (recommended) - uses instance_exec for better mailer access
config.mail_arguments = -> {
  {
    to: params[:pay_customer].email,
    cc: "admin@example.com"
  }
}

# Old syntax (still supported for backward compatibility)
config.mail_arguments = ->(mailer, params) {
  {
    to: params[:pay_customer].email,
    cc: "admin@example.com"
  }
}
```

## Changes Made

1. **`app/mailers/pay/user_mailer.rb`**: Added arity check to support both syntaxes
2. **`lib/pay.rb`**: Updated default configuration to handle `mail_to` arity within `mail_arguments`
3. **`test/pay_test.rb`**: Fixed tests to properly test both syntaxes with instance_exec and explicit arguments
4. **`docs/2_configuration.md`**: Updated documentation to show both syntax examples with clear explanations

## Testing

- Created comprehensive tests for both syntaxes
- Verified backward compatibility with existing configurations
- Tested that instance_exec provides access to mailer instance methods

## Impact

This fix ensures that:
- Users following the current documentation won't encounter runtime errors
- Existing configurations using the old syntax continue to work
- New configurations can use the cleaner instance_exec syntax

## Verification Script

I've tested this fix with a verification script that confirms both syntaxes work correctly:

```ruby
# Test results:
✓ PASS: Instance exec syntax (arity 0) works correctly
✓ PASS: Explicit arguments syntax (arity 2) works correctly  
✓ PASS: mail_to with arity 0 works correctly
✓ PASS: mail_to with explicit arguments works correctly
```

Fixes the issue where users following the official documentation would encounter runtime errors when Pay attempts to send emails (receipts, refunds, subscription renewals, etc.).

### DIFF
--- a/app/mailers/pay/user_mailer.rb
+++ b/app/mailers/pay/user_mailer.rb
@@ -35,7 +35,12 @@ module Pay
     private
 
     def mail_arguments
-      instance_exec(&Pay.mail_arguments)
+      # Support both instance_exec (arity 0) and explicit arguments (arity 2) for backward compatibility
+      if Pay.mail_arguments.arity == 0
+        instance_exec(&Pay.mail_arguments)
+      else
+        Pay.mail_arguments.call(self, params)
+      end
     end
   end
 end
--- a/docs/2_configuration.md
+++ b/docs/2_configuration.md
@@ -159,13 +159,27 @@ Pay.setup do |config|
   config.emails.subscription_trial_will_end = true
   config.emails.subscription_trial_ended = true
 
-  # Customize who receives emails. Useful when adding additional recipients other than the Pay::Customer. This defaults to the pay customer's email address.
+  # Customize who receives emails. Useful when adding additional recipients other than the Pay::Customer. 
+  # This defaults to the pay customer's email address.
+  # The lambda is executed with instance_exec in the mailer context, so you have access to params and mailer methods.
+  # config.mail_to = -> { "#{params[:pay_customer].customer_name} <#{params[:pay_customer].email}>" }
+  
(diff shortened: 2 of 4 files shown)


marckohlbrugge:
@excid3 I think the tests need updating too https://github.com/pay-rails/pay/blob/5160efb9e77a16a4a89d26f557f4e41f623387f7/test/pay_test.rb#L184-L190

PR #1195, opened 2025-09-19
Add unpaid

matiascubero:


excid3:
Pay already has this. https://github.com/pay-rails/pay/blob/main/app/models/pay/subscription.rb#L108

PR #1199, opened 2025-10-03
Fixed issue with the Paddle Classic Signature Verifier

deanpcmad:
On newer versions of OpenSSL, the original code fails with a `OpenSSL::PKey::PKeyError: EVP_DigestVerifyInit: invalid digest` error. This PR updates it so it uses the new style of verification.

### DIFF
--- a/lib/pay/paddle_classic/webhooks/signature_verifier.rb
+++ b/lib/pay/paddle_classic/webhooks/signature_verifier.rb
@@ -38,9 +38,24 @@ module Pay
           data_serialized = serialize(data_sorted, true)
 
           # verify the data
-          digest = OpenSSL::Digest.new("SHA1")
           pub_key = OpenSSL::PKey::RSA.new(public_key)
-          pub_key.verify(digest, signature, data_serialized)
+
+          begin
+            # Try the old verification method first (works on older OpenSSL versions)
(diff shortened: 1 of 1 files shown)


excid3:
Their [docs](https://developer.paddle.com/classic/webhook-reference/ZG9jOjI1MzUzOTg2-verifying-webhooks) still show using:

```ruby
digest    = OpenSSL::Digest::SHA1.new
pub_key   = OpenSSL::PKey::RSA.new(public_key).public_key
verified  = pub_key.verify(digest, signature, data_serialized)
```

This doesn't work anymore? Do you know what the change in OpenSSL was?

If we know which version of OpenSSL is different, we can check the version number instead of using rescues.

excid3:
Does the new version work with an older OpenSSL? If so we could just use that across both versions.

deanpcmad:
Well I'm unsure on the best way then, because it depends on the version of OpenSSL that someone would have installed. Most systems would have higher than v3 so we could remove the initial check and just use the updated version?

excid3:
I wasn't able to reproduce this with macOS Tahoe which had the other OpenSSL issues and the latest OpenSSL 4.0 gem so I'm going to close this because it might have been related to those changes which are now fixed.

I did update the verifier to use the matching code for SHA1 that is in their docs though.

PR #1225, opened 2026-04-09
Update Stripe `retry_failed_payment` api access to payment intent

mguidetti:
## Pull Request

**Summary:**
Fix `retry_failed_payment` for Stripe API versions > 2025-03-31

**Related Issue:**

**Description:**
Stripe API 2025-03-31 added support for multiple payments to invoices and removed the `payment_intent` field from the Invoice object in favor of the payments collection (https://docs.stripe.com/changelog/basil/2025-03-31/add-support-for-multiple-partial-payments-on-invoices)

`retry_failed_payment` was referencing `invoice.payment_intent`, which raises a `NoMethodError` on Stripe API versions >= 2025-03-31. 

This PR changes `retry_failed_payment` to access the payment intent through the `payments` collection instead. It selects the first payment using `.first`, so this doesn't actually support multiple payments like the API does, but perhaps this is fine for this gem's current purposes? 

**Testing:**

I didn't find any existing tests for this method and was unsure the best way to test it. 

**Checklist:**

- [x] All existing tests pass

**Additional Notes:**

### DIFF
--- a/app/models/pay/stripe/subscription.rb
+++ b/app/models/pay/stripe/subscription.rb
@@ -334,7 +334,7 @@ module Pay
 
       # Retries the latest invoice for a Past Due subscription and attempts to pay it
       def retry_failed_payment(payment_intent_id: nil)
-        payment_intent_id ||= api_record.latest_invoice.payment_intent.id
+        payment_intent_id ||= api_record.latest_invoice.payments.first.payment.payment_intent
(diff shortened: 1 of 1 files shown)


excid3:
That should be fine, as presumably the first payment failed and just needs to be retried. I don't know that many people are using these methods, so we may also want to just deprecate them and remove them at some point.

PR #1228, opened 2026-05-09
feat: touch customer and owner on subscription changes

luizkowalski:
## Pull Request

**Summary:**
I have the following code on my app:

```ruby
def subscription
  @subscription ||= Subscription.new(self)
end
```

Where `Subscription` is a PORO that handles everything subscription-related (status, when it's the next charge, whether the user is subscribed, etc)

There are a lot of places where I need to check for the user's subscription status so I figured that this rareley changes, I might cache the status and avoid paying the database query toll

```ruby
def subscribed?
  Rails.cache.fetch("subscription_for_#{@user.cache_key_with_version}", expires_in: 1.hour) do
    @user.payment_processor.on_trial_or_subscribed?(name: NAME)
  end
end
```

The thing is that when the subscription status changes or the customer model is updated, the owner (in my case, `User`) isn't `touch`ed, so the app still thinks they are either subscribed or not, depending on the previous state.

This PR touches the customer and the owner whenever the subscription is updated, causing the cache key to change and allowing the subscription to be cached safely

**Testing:**
Apart from the added tests, I tested locally in my app and managed to safely bust the cache when the user cancels or resubscribes.

**Checklist:**

- [X] Tests have been added or updated to cover the changes
- [X] All existing tests pass

**Additional Notes:**
I'm not sure about the implications of this change, if this would break other people and app's expectations but honestly, I thought this was already a thing

### DIFF
--- a/app/models/pay/customer.rb
+++ b/app/models/pay/customer.rb
@@ -1,6 +1,6 @@
 module Pay
   class Customer < Pay::ApplicationRecord
-    belongs_to :owner, polymorphic: true
+    belongs_to :owner, polymorphic: true, touch: true
     has_many :charges, dependent: :destroy
     has_many :subscriptions, dependent: :destroy
     has_many :payment_methods, dependent: :destroy
--- a/app/models/pay/customer.rb
+++ b/app/models/pay/customer.rb
@@ -1,6 +1,6 @@
 module Pay
   class Customer < Pay::ApplicationRecord
-    belongs_to :owner, polymorphic: true, touch: true
+    belongs_to :owner, polymorphic: true
     has_many :charges, dependent: :destroy
     has_many :subscriptions, dependent: :destroy
     has_many :payment_methods, dependent: :destroy
--- a/app/models/pay/subscription.rb
+++ b/app/models/pay/subscription.rb
@@ -3,7 +3,7 @@ module Pay
     STATUSES = %w[incomplete incomplete_expired trialing active past_due canceled unpaid paused]
 
     # Associations
-    belongs_to :customer
+    belongs_to :customer, touch: true
     belongs_to :payment_method, optional: true, primary_key: :processor_id
     has_many :charges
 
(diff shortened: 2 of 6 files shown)


excid3:
I think it could make sense to touch the pay customer, but I'm not sure about the owner.

You should also use the ActiveRecord object in the cache key to make this cleaner (and not need this change at all):

```ruby
subscription = @user.payment_processor.subscription(name: NAME)
Rails.cache.fetch(subscription) { subscription.active? }
```

By using the object, you can also eliminate the `expires_in` because the cache key will include the `ID` and `updated_at` of the record and you can let your cache automatically clean up the stale data.

luizkowalski:
`@user.payment_processor.subscription(name: NAME)` doesn't this make a query regardless? so in this case, it will query the subscriptions just to compute the cache key

luizkowalski:
yeah, just double-checked if doing that would result in more queries:

currently:
```
Completed 200 OK in 69ms (Views: 10.6ms | ActiveRecord: 2.6ms (6 queries, 0 cached) | GC: 1.2ms)
```

with `@user.payment_processor.subscription(name: NAME)` as cache key:

```
Completed 200 OK in 146ms (Views: 51.8ms | ActiveRecord: 17.9ms (8 queries, 0 cached) | GC: 5.2ms)
```

it does these two extra queries to fetch the customer and the actual subscription to calculate the cache key, making the cache useless

luizkowalski:
hey chris, any chance to look into it again?

excid3:
I think I'm fine adding the `touch` to the Pay::Customer, but not to the associated `owner`. I don't want to touch outside models if we can help it.

I don't think that'll fully eliminate the queries since you will still need to query the `Pay::Customer`.

Instead, you could just override the `belongs_to` on the models when they're loaded to add the touch option:

```ruby
ActiveSupport.on_load :pay_customer do
  belongs_to :owner, polymorphic: true, touch: true
end

ActiveSupport.on_load :pay_subscription do
  belongs_to :customer, touch: true
end
```

luizkowalski:
updated!

PR #1236, opened 2026-07-28
Add Stripe Organization API key support via optional context credential

rameerez:
## Motivation

[Stripe Organizations](https://docs.stripe.com/get-started/account/orgs) let one business run many Stripe accounts (one per product), and an [organization-level API key](https://docs.stripe.com/keys/organization-api-keys) (`sk_org_…`) can act on any account in the org — Stripe just requires every request to name the target account via the `Stripe-Context` header. stripe-ruby (≥ the v19 line Pay already requires) supports this as global configuration.

Today, using an org key with Pay requires wiring `Stripe.config.stripe_context` yourself in an initializer. This PR makes it a first-class, opt-in credential.

## What this does

- New optional `Pay::Stripe.context` credential — resolved exactly like the existing Stripe credentials (`ENV["STRIPE_CONTEXT"]` → env-scoped → unscoped), applied during `Pay::Stripe.setup`.
- **No behavior change for regular account keys**: with no `context` credential, nothing is set.
- Docs section in `docs/stripe/2_credentials.md` + CHANGELOG entry.

One implementation note: the assignment goes through `::Stripe.config.stripe_context` because the `Stripe` module delegates `stripe_account` but not `stripe_context`.

## Verification

- Two new tests in `test/pay/stripe/processor_test.rb` (unset by default; applied when the credential exists): **2 runs, 2 assertions, 0 failures**.
- Validated in a production Rails app: livemode customer create/delete through stripe-ruby with an org key + context, and Pay's full flow running against the routed account.
- (One pre-existing, unrelated test errors locally for me — `env ignores Stripe credentials when not defined` via `Rails.stub` — untouched by this change.)

Marked draft to invite maintainer direction — happy to adjust naming (`context` vs `stripe_context`), docs placement, or approach.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01PqRjZSshq6872LhXvTBaY7

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Unreleased
 
+* Add support for Stripe Organization API keys: optional `context` credential (or `STRIPE_CONTEXT`) sets `Stripe-Context` so a single `sk_org_...` key can serve as `private_key`
+
 ### 11.6.2
 
 * Use secure compare for Paddle Billing webhook verification
--- a/docs/stripe/2_credentials.md
+++ b/docs/stripe/2_credentials.md
@@ -10,6 +10,21 @@ You can create (or find) your Stripe private (secret) and public (publishable) k
 >
 > By default we're linking to the "test mode" page for API keys so you can get up and running in development. When you're ready to deploy to production, you'll have to toggle the "test mode" option off and repeat all steps again for live payments.
 
+### Organization API keys (optional)
+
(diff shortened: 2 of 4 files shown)


excid3:
Thanks @rameerez 👍 

I didn't realize Stripe had organization keys. I'm sure this will come in handy for someone.

PR #1248, opened 2026-08-12
Preserve an existing Stripe API key during setup

manuelmeurer:
## Pull Request

**Summary:**

Preserve an API key configured directly through `stripe-ruby`, while continuing to use Pay’s configured private key as a fallback.

**Related Issue:**

Related to #456.

**Description:**

`Pay::Stripe.setup` currently assigns `Stripe.api_key = private_key` unconditionally. This can overwrite an API key already configured by the host application, including replacing it with `nil` when Pay does not have its own Stripe credentials.

Changing the assignment to `Stripe.api_key ||= private_key` allows applications to configure `stripe-ruby` directly without changing Pay’s existing fallback behavior.

**Testing:**

Added regression tests verifying that:

- Pay’s configured private key is used when `Stripe.api_key` is unset.
- An existing `Stripe.api_key` is preserved.

Commands run:

- `bundle exec ruby -Itest test/pay/stripe/processor_test.rb -n /setup/`
- `bundle exec standardrb lib/pay/stripe.rb test/pay/stripe/processor_test.rb`

The focused setup tests pass with 4 tests and 4 assertions. The complete processor test file currently encounters an unrelated existing Rails 8.1 compatibility error involving `Rails.stub`.

Not applicable.

**Checklist:**

- [x] Documentation has been updated (not applicable)
- [ ] All existing tests pass

### DIFF
--- a/lib/pay/stripe.rb
+++ b/lib/pay/stripe.rb
@@ -41,7 +41,7 @@ module Pay
     end
 
     def self.setup
-      ::Stripe.api_key = private_key
+      ::Stripe.api_key ||= private_key
 
       # When private_key is a Stripe Organizations API key (sk_org_...),
       # every request must identify the target account with the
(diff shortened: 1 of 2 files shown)


excid3:
I've thought about this too and just never remembered to do it. Thanks @manuelmeurer 👍

PR #1251, opened 2026-09-01
Add Square payment processor (card payments, card-on-file, refunds)

Willardgmoore:
## What this adds

A **Square** payment processor covering the payment/card/refund slice, following the existing processor structure (`Pay::Square::Customer` / `PaymentMethod` / `Charge`, `Pay::Webhooks::Square::*`, `lib/pay/square.rb`, `docs/square/*`).

The use case: charge on behalf of a tenant's **own connected Square account** (not merchant of record) with an optional application fee.

> 🙏 **I'm opening this early specifically to get your feedback** — especially on the per-merchant credential approach and the `square_account` naming (below). Heads up: we have **not fully wired this into our own app yet**, so I'm keeping it a draft. We'll be consuming it shortly and will confirm everything is clean and functioning end-to-end, then mark it ready. Happy to adjust anything to match how you'd like it built.

## The one novel design point (feedback wanted)

Unlike Stripe Connect (one platform key + a `stripe_account` request header), Square issues a **distinct OAuth access token per merchant**. There's no shared key, so this cannot reuse the `stripe_context`-style mechanism from #1236 (that's a single-key + selector model). Instead:

- A configurable resolver — `Pay::Square.access_token_provider = ->(pay_customer, force_refresh = false) { ... }` — returns a fresh per-merchant token.
- The processor builds a `Square::Client` **per operation** from that token (a first for the gem; every existing processor sets one global SDK key).
- On a `401` it calls the resolver once with `force_refresh: true` and retries once (the idempotency key is held stable across the retry so it can't double-charge).

The app owns token storage/refresh (tokens expire ~30 days); Pay only asks for one when it needs it. Does this shape work for you, or would you prefer a different seam?

## `square_account` (feedback wanted)

To keep customers isolated per merchant, I added a `square_account` column (customers/charges/payment_methods) and included it in the `set/add_payment_processor` dedup key — mirroring what you did for `stripe_account` in #1198. Kept it processor-specific rather than introducing a generic `account:` param, to match precedent. If you'd rather generalize that, I'm glad to.

## Scope / boundary

In scope: create/find customer, card-on-file (Cards API), charge (`CreatePayment` + `app_fee_money`), refunds, and payment/refund/card/customer webhooks.

Out of scope (deliberately): Square OAuth authorize/callback, token refresh, payouts/reconciliation, catalog, and checkout UI — all app-side. No `Subscription`/`Merchant` classes (Square subscriptions are a separate effort; merchant onboarding is app-side OAuth).

## Notes

- Webhooks are Pay-mounted (`/pay/webhooks/square`), verified via `x-square-hmacsha256-signature` (HMAC over `notification_url + body`). Handlers **re-fetch** authoritative state rather than trusting the payload, since Square delivery is unordered/at-least-once.
- Square refunds are async (`PENDING → COMPLETED`), so `amount_refunded` is synced from the API rather than incremented optimistically.
- No server-side SCA step in Square (3DS is client-side in the Web Payments SDK), so `Pay::Payment`/`has_incomplete_payment?` don't apply — documented in `docs/square/1_overview.md`.
- SDK: `square.rb ~> 44`. Full suite green; new WebMocked suite under `test/pay/square/` covers customer, charge, payment method, cross-merchant isolation, webhook signature (accept/forge/tamper), the 401 refresh-retry, and a token-never-persisted check.

Thanks for the gem — and for any direction here. 🙏

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Unreleased
 
+* Add Square payment processor (card payments, card-on-file, refunds, webhooks) with per-merchant OAuth via a configurable `Pay::Square.access_token_provider`. Charges on a tenant's own connected Square account with optional `app_fee_money`. Adds a `square_account` column to scope customers per merchant.
+
 ### 11.7.2
 
 * `Pay::Customer#subscription` now prefers an active or paused subscription over a newer canceled one, falling back to the most recently created subscription when none are active
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,7 @@ gem "braintree", ">= 2.92.0"
 gem "lemonsqueezy", "~> 1.0"
 gem "paddle", "~> 2.6"
 gem "stripe", "~> 19.0"
+gem "square.rb", "~> 44"
 
 gem "prawn"
 gem "receipts"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -83,6 +83,22 @@ GEM
       uri (>= 0.13.1)
     addressable (2.9.0)
       public_suffix (>= 2.0.2, < 8.0)
+    apimatic_core (0.3.21)
+      apimatic_core_interfaces (~> 0.2.0)
+      certifi (~> 2018.1, >= 2018.01.18)
+      faraday-multipart (~> 1.0)
+      nokogiri (~> 1.13, >= 1.13.10)
+    apimatic_core_interfaces (0.2.3)
+    apimatic_faraday_client_adapter (0.1.6)
+      apimatic_core_interfaces (~> 0.2.0)
+      certifi (~> 2018.1, >= 2018.01.18)
+      faraday (~> 2.0, >= 2.0.1)
+      faraday-follow_redirects (~> 0.2)
+      faraday-gzip (>= 1, < 4)
+      faraday-http-cache (~> 2.2)
+      faraday-multipart (~> 1.0)
+      faraday-net_http_persistent (~> 2.0)
+      faraday-retry (~> 2.0)
     appraisal (3.0.0.rc1)
       bundler
       rake
@@ -95,6 +111,7 @@ GEM
       builder (>= 3.2.4)
       rexml (>= 3.1.9)
     builder (3.3.0)
+    certifi (2018.01.18)
     cgi (0.5.2)
     concurrent-ruby (1.3.8)
     connection_pool (3.0.2)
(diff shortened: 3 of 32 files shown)


excid3:
I won't be able to maintain any more payment processors, but you're welcome to extract this to a Rubygem that integrates with Pay. 👍

Willardgmoore:
Thanks Chris, that makes sense — I've extracted it to a standalone plugin gem: **https://github.com/Willardgmoore/pay-square** (registers the processor from the outside, no fork of Pay). Appreciate the gem and the quick response! 🙏

excid3:
Heck yeah. If you want, you're welcome to make a PR to add it to the README. And long-term we can add it to the pay org if you would like to keep maintaining it.

I'm planning to extract Braintree and Paddle from the main gem. It doesn't bloat anything having them in the gem since they're lazy loaded, but easier to maintain.

Willardgmoore:
Thanks Chris — answering your note above: I'll open the README PR to add it to the Payment Processors list. 🙏 And great to hear Braintree/Paddle are heading toward separate gems; `pay-square` is built exactly that plugin way.

For anyone finding this later, the gem:

👉 **https://github.com/Willardgmoore/pay-square**

Integrates from the outside (no fork of Pay), like [`pay-asaas`](https://github.com/PedroAugustoRamalhoDuarte/pay-asaas): a Rails engine that self-registers (`/pay/webhooks/square`, `configure_webhooks`/`setup`, `include Pay::Attributes`), STI models, and — the one Square-specific wrinkle vs. Stripe Connect — a **distinct OAuth token per merchant** resolved via a configurable `Pay::Square.access_token_provider` (a `Square::Client` per operation), plus a `square_account` column to isolate customers per merchant. Scope: card payments, card-on-file, refunds, webhooks. Targets Pay `~> 11`. Freshly extracted — I'll get it running in our app and published to RubyGems shortly.

On the **pay org** offer — yes, I'd love that, thank you! I'll cut a published release first so there's a tagged version to move over, then we can bring it under the org. Happy to keep maintaining it.

One open question if you're up for it: the only place the plugin reaches into Pay is `set/add_payment_processor`, which I `prepend` to fold `square_account` into the customer dedup key (the isolation `stripe_account` gives). Since that key is hardcoded, any marketplace-style plugin has to monkeypatch it. Would you be open to a small PR adding a generic extension point there so plugins don't have to prepend? Totally fine to keep it in the plugin otherwise.

Thanks again!

PR #1266, opened 2026-09-20
Share the sync skeleton across Stripe Charge, Subscription, and PaymentMethod

excid3:
## Summary

`Charge.sync`, `Subscription.sync`, and `PaymentMethod.sync` each repeated the same fifteen-line prelude (guard against a missing customer, look up the `Pay::Customer`, default the Connect account) and the same rescue-and-retry block. A new `Pay::Stripe::Sync` module holds `sync_with_retries` and `find_pay_customer`; each sync method keeps only its own retrieve call and its model-specific body.

### Behavior changes

- **Retries re-read from Stripe.** The old methods used `object ||= retrieve(...)` at the method level with a method-level `retry`. Ruby keeps locals across `retry`, so the second attempt reused the first read and only re-ran the database write. Now the retrieve lives inside the retried block, so a retry without a passed-in object gets a fresh read. When the caller passes `object:`, it is reused on every attempt and no API call is made, same as before.
- **One delay.** `Subscription.sync` slept a flat 0.1s on retry while the other two used `0.15 * try`. All three now use the growing delay.
- The internal `try:` keyword is gone from all three signatures. Nothing outside these methods passed it. `retries:` stays.
- The debug log lines now say `Stripe charge ch_123` (from `object.object`) instead of `Stripe Charge ch_123`.

### Also in this PR

- `SubscriptionCreated` and `SubscriptionDeleted` inherit from `SubscriptionUpdated`. All three had the identical one-line body. Class names are unchanged, so existing subscribers and overrides keep working. Same pattern as `CheckoutSessionAsyncPaymentSucceeded < CheckoutSessionCompleted`.
- `Pay::Stripe::Customer` used two idioms for "make sure the Stripe customer exists and give me its ID": `api_record unless processor_id?` at seven sites and `processor_id || api_record.id` at five. One private `stripe_customer_id` replaces all twelve. `checkout_charge` also drops its guard, since `checkout` already does it.

Net: 9 files, +202 / −223.

## Test plan

- [x] Three new tests: a retry without a passed object retrieves twice; a retry with a passed object never retrieves; exhausted retries re-raise.
- [x] Full suite: 584 runs, 0 failures, 0 errors. Every existing sync, webhook, and customer test passes unchanged.
- [x] `standardrb` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


excid3:
Pushed review fixes:

- `return unless (pay_customer = find_pay_customer(...))` replaces `or next` at all three sites.
- `find_pay_customer` is one line. The two debug log branches for a missing customer are gone (noted in the changelog).
- The prelude comments repeated in all three sync methods are dropped; the module doc covers it.
- `add_payment_method` uses `stripe_customer_id` throughout instead of mixing it with `processor_id`.
- The retry test now hits the real `RecordInvalid`: an existing row plus a first `find_by` that returns nil (a stale read), so `create!` collides with the uniqueness validation for real and the retry's second `retrieve` is asserted. `create!` is no longer stubbed.

Full suite: 584 runs, 0 failures.

